PR #5, opened 2016-03-21
jQuery removal - closes #2

### LINKED ISSUE #2: Remove jQuery from dependencies
Udash core should not require jQuery to work.

Starzu:
Udash does not depend on jQuery anymore. 

@malcheg, @grzesiul - take a look, please.

### DIFF
--- a/frontend/src/test/scala/io/udash/ApplicationTest.scala
+++ b/frontend/src/test/scala/io/udash/ApplicationTest.scala
@@ -3,6 +3,8 @@ package io.udash
 import io.udash.testing._
 
 class ApplicationTest extends UdashFrontendTest {
+  import scalatags.JsDom.all._
+
   "Application" should {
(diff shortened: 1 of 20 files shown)


frontend/src/test/scala/io/udash/ApplicationTest.scala
```diff
@@ -11,7 +13,7 @@ class ApplicationTest extends UdashFrontendTest {
     val vpRegistry: TestViewPresenterRegistry = new TestViewPresenterRegistry(Map.empty, viewPresenter)
     val app = new Application[TestState](routing, vpRegistry, RootState, urlProvider)
 
-    app.run(null)
+    app.run(div().render)
```
grzesiul: All replacements of "div().render" have meaning of "empty component", so it can be wrapped in explicit method.

malcheg:
It looks good

PR #6, opened 2016-04-04
Fixed ViewRenderer problem with adding more than one view at once.

Starzu:
@malcheg - take a look, please.


malcheg:
I think now it's ok .

PR #13, opened 2016-05-02
Fixed filter on a transformed property

Starzu:
@grzesiul  - take a look, please.

### DIFF
--- a/core/frontend/src/main/scala/io/udash/properties/SeqProperty.scala
+++ b/core/frontend/src/main/scala/io/udash/properties/SeqProperty.scala
@@ -141,17 +141,17 @@ class FilteredSeqProperty[A, ElemType <: ReadableProperty[A]]
 
   origin.elemProperties.foreach(p => registrations(p) = p.listen(elementChanged(p)))
   origin.listenStructure(patch => {
-    patch.removed.foreach(p => {
+    patch.removed.foreach(p => if (registrations.contains(p)) {
       registrations(p).cancel()
       registrations.remove(p)
     })
     patch.added.foreach(p => registrations(p) = p.listen(elementChanged(p)))
 
     val added = patch.added.filter(p => matcher(p.get))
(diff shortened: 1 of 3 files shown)


core/frontend/src/main/scala/io/udash/properties/SeqProperty.scala
```diff
@@ -141,17 +141,17 @@ class FilteredSeqProperty[A, ElemType <: ReadableProperty[A]]
...
     })
     patch.added.foreach(p => registrations(p) = p.listen(elementChanged(p)))
 
     val added = patch.added.filter(p => matcher(p.get))
-    val removed = patch.removed.filter(p => filteredProps.contains(p) && matcher(filteredValues(filteredProps.indexOf(p))))
+    val removed = patch.removed.filter(p => matcher(p.get))
     if (added.nonEmpty || removed.nonEmpty) {
       val props = loadPropsFromOrigin()
-      val idx = if (added.nonEmpty) props.indexOf(added.head) else filteredProps.indexOf(removed.head)
+      val idx = origin.elemProperties.slice(0, patch.idx).count(p => matcher(p.get))
```
grzesiul: Quite surprising - idx looks like index, so indicates previous version, but you obtain it *counting* items matching predicate, am I mi ssing someting?

grzesiul:
lgtm besides that one doubt

core/frontend/src/main/scala/io/udash/properties/SeqProperty.scala
Starzu: Yes, it's surprising. :) As you can see in tests, the problem was related to `SeqProperty -> transform -> filter` chain. When you call `SeqProperty::get` on transformed property it will create Seq with new objects every time (it might be slower, but transformations don't require additional memory this way) and that is why we cannot use `indexOf` method here.

grzesiul:
lgtm, plz merge

PR #14, opened 2016-05-04
Automatic GenCodecs in RPC

Starzu:
@ghik - take a look, please.

### DIFF
--- a/i18n/shared/src/main/scala/io/udash/i18n/types.scala
+++ b/i18n/shared/src/main/scala/io/udash/i18n/types.scala
@@ -1,24 +1,8 @@
 package io.udash.i18n
 
-import com.avsystem.commons.serialization.GenCodec
+import com.avsystem.commons.serialization.transparent
 
-case class Lang(lang: String) extends AnyVal
-case class BundleHash(hash: String) extends AnyVal
+@transparent case class Lang(lang: String) extends AnyVal
+@transparent case class BundleHash(hash: String) extends AnyVal
 case class Bundle(hash: BundleHash, translations: Map[String, String])
-case class Translated(string: String) extends AnyVal
-
-object Lang {
-  implicit val codec = GenCodec.auto[Lang]
-}
-
-object BundleHash {
-  implicit val codec = GenCodec.auto[BundleHash]
-}
-
-object Bundle {
-  implicit val codec = GenCodec.auto[Bundle]
-}
-
-object Translated {
-  implicit val codec = GenCodec.auto[Translated]
-}
+@transparent case class Translated(string: String) extends AnyVal
--- a/rpc/shared/shared/src/main/scala/io/udash/rpc/AutoUdashRPCFramework.scala
+++ b/rpc/shared/shared/src/main/scala/io/udash/rpc/AutoUdashRPCFramework.scala
@@ -1,0 +1,15 @@
+package io.udash.rpc
+
+import com.avsystem.commons.serialization._
+
+import scala.language.postfixOps
+
+trait AutoUdashRPCFramework extends UdashRPCFramework {
+  type RawValue = String
+
+  val RawValueCodec = implicitly[GenCodec[String]]
+
+  def stringToRaw(string: String): RawValue = string
+  def rawToString(raw: RawValue): String = raw
+}
+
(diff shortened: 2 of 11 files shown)


rpc/shared/shared/src/main/scala/io/udash/rpc/AutoUdashRPCFramework.scala
```diff
@@ -0,0 +1,16 @@
+package io.udash.rpc
+
+import com.avsystem.commons.serialization._
+
+import scala.language.postfixOps
+
+trait AutoUdashRPCFramework extends UdashRPCFramework {
+  type RawValue = String
+
+  implicit val RawValueCodec: GenCodec.Auto[RawValue] =
```
ghik: Why is this needed? `GenCodec.Auto` always tries to wrap already existing `GenCodec`.

rpc/shared/shared/src/test/scala/io/udash/rpc/UPickleUdashRPCFramework.scala
```diff
@@ -135,7 +135,7 @@ object UPickleUdashRPCFramework extends UdashRPCFramework {
     }
   }
 
-  implicit val RawValueCodec: GenCodec[RawValue] = new GenCodec[Js.Value] {
+  implicit val RawValueCodec: GenCodec.Auto[RawValue] = GenCodec.Auto(new GenCodec[Js.Value] {
```
ghik: Same here - ideally you shouldn't ever need to declare any implicit of type `GenCodec.Auto`. Instead, declare a `GenCodec` and if you ask for `GenCodec.Auto`, it will wrap the existing `GenCodec`.

I mean, this should work that way - if it doesn't, let me know.

ghik:
General remark: if you had an explicitly declared `GenCodec`, it's totally fine to leave it - it avoids (byte)code duplication and helps with incremental compilation. There should be no need to declare any implicits of type `GenCodec.Auto`.

rpc/shared/shared/src/main/scala/io/udash/rpc/AutoUdashRPCFramework.scala
```diff
@@ -0,0 +1,16 @@
+package io.udash.rpc
+
+import com.avsystem.commons.serialization._
+
+import scala.language.postfixOps
+
+trait AutoUdashRPCFramework extends UdashRPCFramework {
+  type RawValue = String
+
+  implicit val RawValueCodec: GenCodec[RawValue] =
```
ghik: This is minor, but I think it's a problem I've run into at least a few times and I think there is slightly better approach to this:

If base trait needs an implicit, don't declare its abstract method as `implicit`. Instead, create a private method that delegates to abstract method and make the private one `implicit`.
Why? Because you could then do this:

```scala
val RawValueCodec = GenCodec[String]
```
and you don't need to point explicitly to `GenCodec.StringCodec`.

rpc/shared/shared/src/test/scala/io/udash/rpc/RpcMessagesTest.scala
```diff
@@ -121,15 +121,15 @@ trait RpcMessagesTestScenarios extends UdashSharedTest with Utils {
         override def read(input: Input): TwoItems = {
           val obj = input.readObject().get
           obj.nextField()._2.skip()
-          val i2 = codecCI.read(obj.nextField()._2)
+          val i2 = implicitly[GenCodec[CompleteItem]].read(obj.nextField()._2)
```
ghik: just `GenCodec[CompleteItem]`

rpc/shared/shared/src/test/scala/io/udash/rpc/RpcMessagesTest.scala
```diff
@@ -121,15 +121,15 @@ trait RpcMessagesTestScenarios extends UdashSharedTest with Utils {
...
           TwoItems(null, i2)
         }
 
 
         override def write(output: Output, value: TwoItems): Unit = {
           val obj = output.writeObject()
-          codecCI.write(obj.writeField("i1"), value.i1)
-          codecCI.write(obj.writeField("i2"), value.i2)
+          implicitly[GenCodec[CompleteItem]].write(obj.writeField("i1"), value.i1)
```
ghik: same here

i18n/shared/src/main/scala/io/udash/i18n/types.scala
```diff
@@ -1,24 +1,6 @@
 package io.udash.i18n
 
-import com.avsystem.commons.serialization.GenCodec
-
 case class Lang(lang: String) extends AnyVal
```
ghik: Very minor, but you may want to take a look at `@transparent` annotation for `GenCodec`

ghik:
LGTM

PR #15, opened 2016-05-09
User activity mixins

ddworak:
Additional mixins for tracking:
* app navigation
* endpoint calls

The only _breaking_ change is in `RoutingRegistry` (previous state passed on matching URLs). but it might prove useful in other use cases as well.

### DIFF
--- a/rpc/backend/src/test/scala/io/udash/rpc/internals/ExposesServerRPCTest.scala
+++ b/rpc/backend/src/test/scala/io/udash/rpc/internals/ExposesServerRPCTest.scala
@@ -125,10 +125,46 @@ class ExposesServerRPCTest extends UdashBackendTest {
   "DefaultExposesServerRPC" should tests(createDefaultRpc)
   "CustomExposesServerRPC" should tests(createCustomRpc)
   "LoggingExposesServerRPC" should {
-    tests(createLoggingRpc)
     import io.udash.rpc.InnerRPC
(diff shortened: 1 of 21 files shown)


rpc/backend/src/test/scala/io/udash/rpc/internals/ExposesServerRPCTest.scala
```diff
@@ -98,10 +100,35 @@ class ExposesServerRPCTest extends UdashBackendTest {
...
+      override protected def localRpcAsRaw: framework.AsRawRPC[TestRPC] = framework.AsRawRPC[TestRPC]
+
+      override def log(rpcName: String, methodName: String, args: Seq[String]): Unit = loggedCalls += s"$rpcName $methodName $args"
+    }
   }
 
   "DefaultExposesServerRPC" should tests(createDefaultRpc)
   "CustomExposesServerRPC" should tests(createCustomRpc)
+  "LoggingExposesServerRPC" should {
+    tests(createLoggingRpc)
```
Starzu: Usage of tests() in this case is not a good idea, I think it would be better to create separate calls for annotated and unannotated methods here.

Starzu:
Looks good except this one test. ;) Rebase and squash the commits, please.

ddworak:
done

Starzu:
LGTM

PR #27, opened 2016-05-16
Improvements in the Atmosphere RPC implementation

Starzu:
* WebSocket works in both directions and properly waits for a reconnection
* if WS is not available, it works for the client->server communication with returning values from the server to the client (server pushes are not supported)

### DIFF
--- a/core/frontend/src/main/scala/io/udash/utils/Logger.scala
+++ b/core/frontend/src/main/scala/io/udash/utils/Logger.scala
@@ -30,7 +30,7 @@ private object ConsoleLogger extends Logger {
 }
 
 /**
-  * Provides `logger` reference to [[io.udash.utils.ConsoleLogger]].
+  * Provides `logger` reference to io.udash.utils.ConsoleLogger.
   */
 trait StrictLogging {
   val logger: Logger = ConsoleLogger
--- a/rpc/backend/src/main/scala/io/udash/rpc/AtmosphereService.scala
+++ b/rpc/backend/src/main/scala/io/udash/rpc/AtmosphereService.scala
@@ -1,15 +1,19 @@
 package io.udash.rpc
 
+import java.util.UUID
 import javax.servlet.ServletInputStream
 import javax.servlet.http.HttpServletResponse
 
 import com.typesafe.scalalogging.LazyLogging
 import io.udash.rpc.internals._
 import org.atmosphere.cpr.AtmosphereResource.TRANSPORT
-import org.atmosphere.cpr.{AtmosphereConfig, AtmosphereResource, AtmosphereResourceEvent, AtmosphereServletProcessor}
+import org.atmosphere.cpr._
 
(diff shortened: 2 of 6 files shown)


rpc/frontend/src/main/scala/io/udash/rpc/internals/ServerConnector.scala
```diff
@@ -27,38 +27,55 @@ abstract class AtmosphereServerConnector[RPCRequest](private val serverUrl: Stri
...
+        onError = (res: AtmosphereResponse) => ready(false),
+        onClose = (res: AtmosphereResponse) => ready(false),
+        onClientTimeout = (res: AtmosphereResponse) => ready(false),
+        onMessage = (res: AtmosphereResponse) => handleMessage(res.responseBody),
+        onMessagePublished = (req: AtmosphereRequest, res: AtmosphereResponse) => handleMessage(req.responseBody)
+      )
+      Atmosphere.subscribe(atmRequest)
+    } else {
+      isReady = true
+      val atmRequest: AtmosphereRequest = AtmosphereRequest(
```
ddworak: Extract common settings between these two branches and reuse, this is hard to read and even harder to maintain.

rpc/backend/src/main/scala/io/udash/rpc/AtmosphereService.scala
```diff
@@ -38,20 +42,27 @@ trait AtmosphereServiceConfig[ServerRPCType] {
...
   extends AtmosphereServletProcessor with LazyLogging {
 
-  override def init(config: AtmosphereConfig): Unit =
-    BroadcastManager.init(config.getBroadcasterFactory, config.metaBroadcaster())
+  private var brodcasterFactory: BroadcasterFactory = _
 
-  override def onRequest(resource: AtmosphereResource): Unit = {
-    config.initRpc(resource)
+  override def init(config: AtmosphereConfig): Unit = {
+    brodcasterFactory = config.getBroadcasterFactory
```
ddworak: Do we ever want to re-initialize this? Make it a `val` and init in constructor unless we do.

Starzu: This method is called from Atmosphere internals, so we cannot pass the `BroadcasterFactory` as constructor argument.

core/frontend/src/main/scala/io/udash/utils/Logger.scala
```diff
@@ -30,7 +30,7 @@ private object ConsoleLogger extends Logger {
 }
 
 /**
-  * Provides `logger` reference to [[io.udash.utils.ConsoleLogger]].
+  * Provides `logger` reference to io.udash.utils.ConsoleLogger.
```
ddworak: What's the reason for removing the doc link?

Starzu: `Could not find any member to link for "io.udash.utils.ConsoleLogger"` - it's private object.

PR #28, opened 2016-05-17
Frontend UrlLogging changes

Starzu:
What do you think about such change of  `UrlLogging`? I think it's better because `matchUrl` method should be responsible only for matching `Url` to `State` and it does not need previous state for this.


ddworak:
I'm not sure about that lack of need for previous state for matching the URL to new one. I actually thought that someone might find that additional information useful (e.g. if the client has just entered the website with some specific URL we might want to modify the state we're matching him to smh). 
Nevertheless, if we don't want the users to do that, the code looks good to me.

Starzu:
I don't like the idea of matching an URL to a state based on the previous state. I definitely don't want users to do that.

PR #30, opened 2016-05-20
Input debouncing

Starzu:

### DIFF
--- a/core/frontend/src/main/scala/io/udash/bindings/TextArea.scala
+++ b/core/frontend/src/main/scala/io/udash/bindings/TextArea.scala
@@ -4,7 +4,8 @@ import io.udash.bindings.modifiers.TextInputsModifier
 import io.udash.properties.Property
 import org.scalajs.dom.{html, _}
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.{Duration, DurationInt}
+import scala.language.postfixOps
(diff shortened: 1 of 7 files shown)


core/frontend/src/main/scala/io/udash/bindings/TextArea.scala
```diff
@@ -40,4 +42,12 @@ object TextArea {
...
     textarea(bind, xs)
   }
+
+  /**
+    * @param property Property to bind.
+    * @param xs Additional Modifiers, don't use modifiers on value, onchange and onkeyup attributes.
+    * @return HTML textarea with bound Property, applied modifiers and nested options.
+    */
+  def apply(property: Property[String], xs: Modifier*): JsDom.TypedTag[html.TextArea] =
+    apply(property, 20 millis, xs:_*)
```
ddworak: I don't like the debounce enabled by default nor would I expect it after reading this method's doc. How about making this more explicit (maybe `TextArea.debounced` instead of `.apply`)?

Starzu: Take a look, again. Now `apply(property: Property[String], xs: Modifier*)` is deprecated and works exactly the same as before changes. There is  `debounced` method which applies default debouncing.

ddworak:
lgtm

PR #31, opened 2016-05-27
Fixed problem with sending integer as value of Double from client to server.

Starzu:
### Problem

RPC interface has method like:
```scala
def doSth(d: Double): Unit
````

Client calls:
```scala
serverRPC.doSth(123)
```

Result: 
Server cannot deserialize JSON, because it expects double.


ddworak:
lgtm :shipit:

PR #33, opened 2016-06-01
AVS Commons 16

Starzu:
* Udash @RPC and @RPCName
* AVS Commons 1.16.0
* Separated ServerRPCFramework & ClientRPCFramework
* UdashRPCMacros removed


rpc/shared/shared/src/main/scala/io/udash/rpc/annotations.scala
```diff
@@ -0,0 +1,4 @@
+package io.udash.rpc
+
+class RPC extends com.avsystem.commons.rpc.RPC
```
ghik: Why do you need these?

Starzu: I want these annotations to import with `io.udash.rpc._`. Importing `com.avsystem.commons.rpc.RPC` might be confusing.

ddworak:
lgtm

PR #35, opened 2016-06-13
Udash REST support

Starzu:
@ddworak take a look, please. You can find a demo application [here](https://github.com/UdashFramework/udash-demos/tree/rest-spray-io/rest-spray-io).


project/Dependencies.scala
```diff
@@ -70,4 +72,11 @@ object Dependencies extends Build {
   val rpcBackendTestDeps = Def.setting(Seq(
     "org.scalamock" %% "scalamock-scalatest-support" % scalamockVersion
   ).map(_ % Test))
+
+  val restCrossDeps = Def.setting(Seq(
+    "com.avsystem.commons" %%% "commons-shared" % avsCommonsVersion,
+    "fr.hmil" %%% "scala-http-client" % scalaHttpClientVersion
```
ddworak: This is now https://github.com/hmil/RosHTTP, switch before merge.

rest/shared/src/main/scala/io/udash/rest/DefaultRESTConnector.scala
```diff
@@ -0,0 +1,51 @@
...
+        case 4 => Future.failed(RESTConnector.ClientError(resp.statusCode, resp.body))
+        case 5 => Future.failed(RESTConnector.ServerError(resp.statusCode, resp.body))
+      }
+    })
+  }
+
+  private implicit def methodConverter(method: HTTPMethod): Method = method match {
+    case RESTConnector.GET => Method.GET
+    case RESTConnector.POST => Method.POST
+    case RESTConnector.PATCH => Method("PATCH")
```
ddworak: scalahttp backend does not support PATCH, we have to remember about that once we write REST backend support. We'll probably use a more sophisticated lib there though.

rest/shared/src/main/scala/io/udash/rest/DefaultServerREST.scala
```diff
@@ -0,0 +1,41 @@
...
+import io.udash.rest.internal.{RESTConnector, UsesREST}
+
+import scala.concurrent.ExecutionContext
+
+/** Default REST usage mechanism using [[io.udash.rest.DefaultRESTFramework]]. */
+class DefaultServerREST[ServerRPCType](override protected val connector: RESTConnector)
+                                      (implicit override val remoteRpcAsReal: DefaultRESTFramework.AsRealRPC[ServerRPCType],
+                                       override val rpcMetadata: RPCMetadata[ServerRPCType],
+                                       validRest: DefaultRESTFramework.ValidREST[ServerRPCType],
+                                       ec: ExecutionContext)
```
ddworak: Some of these implicit parameters could be expressed in context bounds.

rest/shared/src/main/scala/io/udash/rest/DefaultServerREST.scala
```diff
@@ -0,0 +1,41 @@
...
+  def rawToURLPart(raw: framework.RawValue): String =
+    stripQuotes(framework.rawToString(raw))
+
+  private def stripQuotes(s: String): String =
+    s.stripPrefix("\"").stripSuffix("\"")
+}
+
+object DefaultServerREST {
+  /** Creates [[io.udash.rest.DefaultServerREST]] with [[io.udash.rest.DefaultRESTConnector]] for provided REST interfaces. */
+  def apply[ServerRPCType](host: String, port: Int, pathPrefix: String = "")
```
ddworak: Same here.

rest/shared/src/main/scala/io/udash/rest/internal/RESTConnector.scala
```diff
@@ -0,0 +1,30 @@
...
+
+object RESTConnector {
+  sealed trait HTTPMethod
+  case object GET extends HTTPMethod
+  case object POST extends HTTPMethod
+  case object PATCH extends HTTPMethod
+  case object PUT extends HTTPMethod
+  case object DELETE extends HTTPMethod
+
+  sealed trait RequestError extends Exception {
```
ddworak: Don't call an Exception an Error :confused:

rest/shared/src/main/scala/io/udash/rest/internal/UsesREST.scala
```diff
@@ -0,0 +1,126 @@
...
+  protected def remoteRpcAsReal: AsRealRPC[ServerRPCType]
+
+  protected val connector: RESTConnector
+
+  /** Transform `RawValue` into String used in HTTP request header. */
+  def rawToHeaderArgument(raw: RawValue): String
+  /** Transform `RawValue` into String used as HTTP request query argument. */
+  def rawToQueryArgument(raw: RawValue): String
+  /** Transform `RawValue` into String used as URL part. */
+  def rawToURLPart(raw: RawValue): String
```
ddworak: Does it make sense for these three methods to be separate or even to use anything else than `framework.rawToString`? Seems rather as a possibility for user bad practices, needs to be at least commented.

rest/shared/src/main/scala/io/udash/rest/internal/UsesREST.scala
```diff
@@ -0,0 +1,126 @@
...
+    var metadata: RPCMetadata[_] = rpcMetadata
+    getterChain.reverse.foreach(inv => {
+      parseInvocation(inv, metadata)
+      metadata = metadata.getterResults(inv.rpcName)
+    })
+    parseInvocation(invocation, metadata)
+
+    val restMethod: HTTPMethod = findRestMethod(invocation, metadata)
+
+    connector.send(s"/${urlBuilder.result().mkString("/")}", restMethod, queryArgsBuilder.result(), headersArgsBuilder.result(), body) onComplete {
```
ddworak: I'd inline restMethod and use named parameters here, this is probably the single most important line in this extension.

rest/shared/src/main/scala/io/udash/rest/internal/UsesREST.scala
Starzu: The default serialization uses them to remove quotes from the serialized `String`. In a request body it should look like `"string"`, but in a URL it should be `string`. In other serialization methods it might be even more useful.

ddworak:
lgtm

PR #41, opened 2016-07-15
Property system improvements (part 1)

Starzu:
* ModelProperty based on simple case classes (case classes contating only vals)
* Lazy initialization of subproperties in ModelProperty (fixes recursive model properties)
* A lot of new tests

@ddworak and @malcheg - take a look, please


malcheg:
@Starzu looks good

ddworak:
I've tried hard, yet I found no basis for a charge against this code. :star:

PR #43, opened 2016-07-20
Property - new methods

Starzu:
* `combine[B, O : ModelValue](property: Property[B])(combiner: (A, B) => O): Property[O]` - creates new `Property[O]` synchronized with the both origin
* `combine[B, O : ModelValue](property: ReadableProperty[B])(combiner: (A, B) => O): ReadableSeqProperty[O, ReadableProperty[O]]` - every element of `this` is combined with `property`
* `transform[B : ModelValue](transformer: A => Seq[B]): ReadableSeqProperty[B, ReadableProperty[B]]` - transformation of `ReadableProperty[A]` to `ReadableSeqProperty[B]`
* `transform[B : ModelValue](transformer: A => Seq[B], revert: Seq[B] => A): SeqProperty[B, Property[B]]` -  transformation of `Property[A]` to `SeqProperty[B]` (synchronized in both directions)

### DIFF
--- a/core/frontend/src/main/scala/io/udash/properties/Property.scala
+++ b/core/frontend/src/main/scala/io/udash/properties/Property.scala
@@ -81,18 +81,21 @@ trait ReadableProperty[A] {
   def transform[B](transformer: A => B): ReadableProperty[B] =
     new TransformedReadableProperty[A, B](this, transformer, PropertyCreator.newID())
 
-  def combine[B, O : ModelValue](property: Property[B])(combiner: (A, B) => O): Property[O] = {
-    val output = Property[O]
+  /** Combines two properties into a new one. Created property will be updated after any change in the origin ones. */
+  def combine[B, O : ModelValue](property: ReadableProperty[B], combinedParent: ReadableProperty[_] = null)(combiner: (A, B) => O): ReadableProperty[O] = {
+    val pc = implicitly[PropertyCreator[O]]
+    val output = pc.newProperty(combinedParent)
+
     def update(x: A, y: B): Unit =
       output.set(combiner(x, y))
 
+    output.setInitValue(combiner(get, property.get))
     listen(x => update(x, property.get))
     property.listen(y => update(get, y))
-    update(get, property.get)
     output
   }
 
-  protected[properties] def parent: Property[_]
+  protected[properties] def parent: ReadableProperty[_]
 
   protected[properties] def fireValueListeners(): Unit = {
     val t = get
@@ -130,7 +133,7 @@ class TransformedReadableProperty[A, B](private val origin: ReadableProperty[A],
   override protected[properties] def fireValueListeners(): Unit =
     origin.fireValueListeners()
 
-  override protected[properties] def parent: Property[_] =
+  override protected[properties] def parent: ReadableProperty[_] =
     origin.parent
 
   override def validate(): Unit =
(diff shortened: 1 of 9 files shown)


core/frontend/src/main/scala/io/udash/properties/Property.scala
```diff
@@ -117,6 +140,74 @@ trait ReadableProperty[A] {
...
+  update(origin.get)
+  origin.listen(update)
+
+  private def structureChanged(patch: Patch[Property[B]]): Unit =
+    structureListeners.foreach(_.apply(patch))
+
+  private def update(v: A): Unit = {
+    val transformed = transformer(v)
+    val current = get
+    val commonBegin = {
```
ddworak: I'm thinking about a more "Scala" way of doing this, I'm not sure it's more readable and didn't test too much. Check my draft, build on it or ignore it:
```scala
def commonIdx(s1: Seq[B], s2: Seq[B]): Int = {
  val (longer, shorter) = if(s1.size > s2.size) (s1, s2) else (s2, s1)
  math.max(0, longer.zipWithIndex.indexWhere { case ((elem, idx)) => idx >= shorter.size || elem != shorter(idx) })
}
val commonBegin = commonIdx(transformed, current)
val commonEnd = commonIdx(transformed.reverse, current.reverse)
```

core/frontend/src/main/scala/io/udash/properties/Property.scala
```diff
@@ -117,6 +140,74 @@ trait ReadableProperty[A] {
...
+
+    val patch = if (transformed.size > current.size) {
+      val added: Seq[CastableProperty[B]] = Seq.fill(transformed.size - current.size)(pc.newProperty(this)(executionContext))
+      children.insertAll(commonBegin, added)
+      Patch(commonBegin, Seq(), added, false)
+    } else if (transformed.size < current.size) {
+      val removed = children.slice(commonBegin, commonBegin + current.size - transformed.size)
+      children.remove(commonBegin, current.size - transformed.size)
+      Patch(commonBegin, removed, Seq(), transformed.isEmpty)
+    } else null
```
ddworak: use None

core/frontend/src/main/scala/io/udash/properties/Property.scala
```diff
@@ -117,6 +140,74 @@ trait ReadableProperty[A] {
...
+        .foreach { case (pv, p) => p.set(pv) }
+      if (patch != null) structureChanged(patch)
+      valueChanged()
+    }
+  }
+
+  override def get: Seq[B] =
+    children.map(_.get)
+
+  override def elemProperties: Seq[ReadableProperty[B]] =
```
ddworak: This implements a read-only view. I believe with this implementation it is possible to pattern match on the elements of returned sequence and modify them as regular Properties by mistake.

core/frontend/src/main/scala/io/udash/properties/Property.scala
Starzu:
```scala
def commonIdx(s1: Iterator[B], s2: Iterator[B], longerSize: Int): Int =
  math.max(longerSize,
    s1.zipAll(s2, null, null).zipWithIndex
      .indexWhere { case (((x, y), idx)) => x != y })

val longerSize = if (transformed.size > current.size) transformed.size else current.size
val commonBegin = commonIdx(transformed.iterator, current.iterator, longerSize)
val commonEnd = commonIdx(transformed.reverseIterator, current.reverseIterator, longerSize)
```
What about this? It wont create reversed copy of both sequences.

ddworak:
lgtm!

PR #45, opened 2016-07-21
Handling getter error in RPC call

Starzu:


rpc/backend/src/main/scala/io/udash/rpc/ExposesServerRPC.scala
```diff
@@ -18,8 +19,13 @@ abstract class ExposesServerRPC[ServerRPCType](local: ServerRPCType) extends Exp
 
   /** Handles RPCCall and returns Future with call result. */
   def handleRpcCall(call: RPCCall): Future[RawValue] = {
-    val receiver = rawLocalRpc.resolveGetterChain(call.gettersChain)
-    receiver.call(call.invocation.rpcName, call.invocation.argLists)
+    try {
+      val receiver = rawLocalRpc.resolveGetterChain(call.gettersChain)
+      receiver.call(call.invocation.rpcName, call.invocation.argLists)
```
ddworak: I've seen this fail on SBT/IDEA compilation clash without any logging. We should consider doing this, just so the user knows it's not his fault nor it is ours :blush:

ddworak:
lgtm

PR #51, opened 2016-07-28
FinalView trait with default renderChild implementation

Starzu:


core/frontend/src/main/scala/io/udash/core/Definitions.scala
```diff
@@ -73,7 +83,7 @@ trait State {
 
 /**
   * The implementation of this trait should be injected to [[io.udash.routing.RoutingEngine]].
-  * It should map properly [[Url]] to [[State]], and otherwise.
+  * It should map properly [[io.udash.core.Url]] to [[io.udash.core.State]], and otherwise.
```
ddworak: `otherwise` is not the right word, maybe:
`It should implement a bidirectional mapping between [[io.udash.core.Url]] and [[io.udash.core.State]]`

PR #52, opened 2016-07-29
Bindings with Seq[Element] and View template returning Modifier

Starzu:
### Changes 

* The `getTemplate` method in `View` now returns `Modifier[dom.Element]` - now it's possible to return a property binding directly:
```scala
class BindingView extends FinalView {
  val property = Property("ABCDE")
  override def getTemplate: Modifier = bind(property)
}
```
* All binding builders return `Seq[Element]`.
* The `Component` trait was useless.


core/frontend/src/main/scala/io/udash/bindings/modifiers/ValueModifier.scala
```diff
@@ -11,18 +11,28 @@ import scalatags.generic._
...
-  def builder: (T => Element)
+  def builder: (T => Seq[dom.Element])
   def checkNull: Boolean
   def listen(callback: T => Unit): Registration
 
   override def applyTo(t: dom.Element): Unit = {
-    var element: Element = null
+    var elements: Seq[Element] = null
 
     def rebuild() = {
```
ddworak: This has a lot in common with `ValidationValueModifier.rebuild` - you could probably extract a common trait with this logic.

core/frontend/src/main/scala/io/udash/view/ViewRenderer.scala
```diff
@@ -32,11 +33,16 @@ private[udash] class ViewRenderer(rootElement: => Element) {
...
     views.appendAll(path)
-    val child: Element = rootView.getTemplate
-    if (endpoint.hasChildNodes()) endpoint.replaceChild(child, endpoint.lastChild)
-    else endpoint.appendChild(child)
+
+    // Clear root element
+    for (_ <- 0 until endpoint.childElementCount)
+      endpoint.removeChild(endpoint.childNodes(0))
+
+    val child: Modifier[Element] = rootView.getTemplate
```
ddworak: Could be inlined.

ddworak:
lgtm

PR #53, opened 2016-08-01
isValid returning null - fixed

Starzu:


core/frontend/src/main/scala/io/udash/properties/single/Property.scala
```diff
@@ -106,11 +106,15 @@ trait ReadableProperty[A] {
...
+      val p = Promise[ValidationResult]
+      validationResult = p.future
       CallbackSequencer.queue(s"${this.id.toString}:fireValidation", () => {
         import Validator._
-        validationResult = Future.sequence(
-          validators.collect { case v => v(this.get) }.toSeq
-        ).foldValidationResult
+        p.completeWith(
+          Future.sequence(
+            validators.collect { case v => v(this.get) }.toSeq
```
ddworak: Any reason to `.collect` instead of `.map`?

Starzu: Nope, fixed. ;)

ddworak:
:+1:

PR #59, opened 2016-08-05
`showIf` binding method

Starzu:


core/frontend/src/main/scala/io/udash/bindings/Bindings.scala
```diff
@@ -42,6 +42,19 @@ trait Bindings {
     new SimplePropertyModifier[T](property, true)
 
   /**
+    * Shows provided DOM elements onyl if property value is `true`.
```
ddworak: only

core/frontend/src/main/scala/io/udash/bindings/Bindings.scala
```diff
@@ -42,6 +42,19 @@ trait Bindings {
...
+    *
+    * @param property Property to check.
+    * @param elements  Elements to show if property value is `true`.
+    * @return Modifier for bounded property.
+    */
+  def showIf(property: ReadableProperty[Boolean])(elements: Seq[dom.Element]) =
+    new PropertyModifier[Boolean](property, {
+      case true => elements
+      case false => Seq(emptyStringNode())
```
ddworak: Wouldn't `Seq.empty` work just as well?

Starzu: Right :)

Starzu:
@ddworak updated, take a look

ddworak:
lgtm

PR #60, opened 2016-08-05
SeqProperty: zip, zipAll, zipWithIndex

Starzu:


core/frontend/src/main/scala/io/udash/properties/seq/SeqProperty.scala
```diff
@@ -99,6 +99,22 @@ trait ReadableSeqProperty[A, +ElemType <: ReadableProperty[A]] extends ReadableP
     new CombinedReadableSeqProperty(this, property, executionContext)
   }
 
+  /** Zips elements from `this` and provided `property` by combining every pair using provided `combiner`. */
+  def zip[B, O : ModelValue](property: ReadableSeqProperty[B, _ <: ReadableProperty[B]])(combiner: (A, B) => O): ReadableSeqProperty[O, ReadableProperty[O]] =
```
ddworak: I believe you could simplify the argument type to:
```scala 
ReadableSeqProperty[B, ReadableProperty[B]]
```
The second type parameter is covariant anyway.

core/frontend/src/main/scala/io/udash/properties/seq/SeqProperty.scala
```diff
@@ -99,6 +99,22 @@ trait ReadableSeqProperty[A, +ElemType <: ReadableProperty[A]] extends ReadableP
...
+    new ZippedReadableSeqProperty(this, property, combiner, executionContext)
+
+  /** Zips elements from `this` and provided `property` by combining every pair using provided `combiner`.
+    * Uses `defaultA` and `defaultB` to fill smaller sequence. */
+  def zipAll[B, O : ModelValue](property: ReadableSeqProperty[B, _ <: ReadableProperty[B]])
```
ddworak: Same here

core/frontend/src/main/scala/io/udash/properties/seq/ZippedSeqProperty.scala
```diff
@@ -0,0 +1,119 @@
...
+                                   p: ReadableSeqProperty[B, ReadableProperty[B]],
+                                   combiner: (A, B) => O, defaultA: ReadableProperty[A], defaultB: ReadableProperty[B],
+                                   override val executionContext: ExecutionContext)
+  extends ZippedReadableSeqProperty(s, p, combiner, executionContext) {
+
+  override protected def update(fromIdx: Int): Unit =
+    appendChildren(s.elemProperties.zipAll(p.elemProperties, defaultA, defaultB).drop(fromIdx))
+}
+
+class ZippedWithIndexReadableSeqProperty[A](s: ReadableSeqProperty[A, ReadableProperty[A]],
```
ddworak: Do we really have to copy all this logic? Implementing zipWithIndex with zip does not seem very bad at first glance (there may be some corner cases, but it's worth checking).
```scala
lazy val zipWithIndex: ReadableSeqProperty[(A, Int), ReadableProperty[(A, Int)]] =
  zip(transform((s: Seq[A]) => s.indices))((_,_))
```

Starzu:
@ddworak I've made some improvements, take a look

ddworak:
lgtm

PR #62, opened 2016-08-10
Rename: transform(A -> Seq[B]) to transformToSeq(A -> Seq[B])

Starzu:
The same name for
```scala
def transform[B](transformer: A => B): ReadableProperty[B]
```
and 
```scala
 def transform[B : ModelValue](transformer: A => Seq[B]): ReadableSeqProperty[B, ReadableProperty[B]]
```
introduced problems with inline lambdas.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,27 @@
-sudo: true
+sudo: required
+dist: trusty
 language: scala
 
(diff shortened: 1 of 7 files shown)


ddworak:
lgtm

PR #66, opened 2016-08-30
Numbers parsing improvements

Starzu:


ddworak:
Missing test case for bad number formats, otherwise lgtm.

PR #72, opened 2016-09-19
Updated to scala-commons 1.18.1, adding explicit typeclass instances

ghik:


i18n/shared/src/main/scala/io/udash/i18n/RemoteTranslationRPC.scala
```diff
@@ -21,5 +22,4 @@ trait RemoteTranslationRPC {
   /** Returns map of translations and bundle hash. If `oldHash` is not outdated, this Future will contain None. */
   def loadTranslationsForLang(lang: Lang, oldHash: BundleHash): Future[Option[Bundle]]
 }
-
-
+object RemoteTranslationRPC extends RPCTypeClasses[DefaultServerUdashRPCFramework.type, RemoteTranslationRPC]
```
Starzu: We cannot assume that someone's using `DefaultServerUdashRPCFramework`. I think we cannot provide `RPCTypeClasses` for `RemoteTranslationRPC` and we can add a note in Udash Guide.

i18n/shared/src/main/scala/io/udash/i18n/types.scala
```diff
@@ -1,8 +1,20 @@
 package io.udash.i18n
 
-import com.avsystem.commons.serialization.transparent
+import com.avsystem.commons.serialization.{GenCodec, transparent}
 
 @transparent case class Lang(lang: String) extends AnyVal
+object Lang {
+  implicit val codec: GenCodec[Lang] = GenCodec.materialize[Lang]
+}
```
Starzu: Why not
```scala
object Lang extends HasGenCodec[Lang]
```
?

i18n/shared/src/main/scala/io/udash/i18n/RemoteTranslationRPC.scala
ghik: This by no means makes any assumption about usage of `DefaultServerUdashRPCFramework`. It simply adds instances for it since it's the framework most likely to be used.

i18n/shared/src/main/scala/io/udash/i18n/types.scala
ghik: Well, `HasGenCodec` has a lot of limitations, so I don't want to encourage it so much...

PR #82, opened 2017-01-27
Translations improvements

Starzu:
* TranslationKey0 GenCodec
* TranslationKeys reduce method
* Untranslatable as subtype of TranslationKey0

### DIFF
--- a/i18n/shared/src/main/scala/io/udash/i18n/TranslationKey.scala
+++ b/i18n/shared/src/main/scala/io/udash/i18n/TranslationKey.scala
@@ -1,6 +1,10 @@
 package io.udash.i18n
 
+import com.avsystem.commons.serialization.{GenCodec, Input, Output}
+
 import scala.concurrent.Future
+import scala.util.Try
+import scala.util.control.NonFatal
 
 sealed trait TranslationKey {
   def key: String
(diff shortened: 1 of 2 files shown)


i18n/shared/src/main/scala/io/udash/i18n/TranslationKey.scala
```diff
@@ -106,4 +142,47 @@ object TranslationKey {
...
+  }
+
+  private[i18n] class Untranslatable(override val key: String) extends TranslationKey0 {
+    override def apply()(implicit provider: TranslationProvider, lang: Lang): Future[Translated] =
+      Future.successful(Translated(key))
+  }
+}
+
+object TranslationKey0 {
+  implicit val codec: GenCodec[TranslationKey0] = new GenCodec[TranslationKey0] {
```
ddworak: You could've used `GenCodec.create`, although I don't think it will ever be different from this.

i18n/shared/src/main/scala/io/udash/i18n/TranslationKey.scala
```diff
@@ -106,4 +142,47 @@ object TranslationKey {
...
+    override def read(input: Input): TranslationKey0 = {
+      try {
+        val data = input.readList()
+        val key = data.nextElement().readString()
+        val items = data.iterator(item => item.readString()).toSeq
+        if (items.nonEmpty) new TranslationKey.ReducedTranslationKey(key, items: _*)
```
ddworak: I suggest `data.skipRemaining()` here.

PR #84, opened 2017-01-27
Serialization of RPC exceptions

Starzu:


rpc/shared/shared/src/main/scala/io/udash/rpc/serialization/ExceptionsRegistry.scala
```diff
@@ -0,0 +1,41 @@
...
+  /** Returns GenCodec for provided class name. */
+  def get[T](name: String): GenCodec[T]
+  /** Returns true, if contains GenCodec for provided class name. */
+  def contains(name: String): Boolean
+  /** Returns name used for class T. */
+  def name[T](ex: T): String
+}
+
+object ExceptionsRegistry {
+  val Default: ExceptionsRegistry = new DefaultExceptionsRegistry
```
ghik: Global mutable object? Uncool.
I would simply use empty registry as the default one.

rpc/shared/shared/src/main/scala/io/udash/rpc/serialization/ExceptionsRegistry.scala
```diff
@@ -0,0 +1,41 @@
+package io.udash.rpc.serialization
+
+import com.avsystem.commons.serialization.GenCodec
+
+import scala.collection.mutable
+import scala.reflect.ClassTag
+
+trait ExceptionsRegistry {
```
ghik: Is there really a need to separate trait and implementation?

rpc/shared/shared/src/main/scala/io/udash/rpc/UdashRPCFramework.scala
```diff
@@ -138,56 +143,72 @@ trait UdashRPCFramework extends GetterRPCFramework with ProcedureRPCFramework wi
...
-            response = RawValueCodec.read(fi)
-          case "callId" =>
-            callId = fi.readString()
-          case "cause" =>
-            cause = fi.readString()
-          case "errorMsg" =>
-            errorMsg = fi.readString()
-          case "type" =>
-            tpe = fi.readString()
+  def RPCResponseCodec(exceptionsRegistry: ExceptionsRegistry): GenCodec[RPCResponse] =
```
ghik: If this codec is not `implicit` then I think other ones should also not be.
Or they shouldn't be codecs at all, but I guess you already explained that you're planning to do this later.

rpc/shared/shared/src/main/scala/io/udash/rpc/UdashRPCFramework.scala
```diff
@@ -138,56 +143,72 @@ trait UdashRPCFramework extends GetterRPCFramework with ProcedureRPCFramework wi
...
+            RawValueCodec.write(obj.writeField("response"), response)
+            obj.writeField("callId").writeString(callId)
+          case RPCResponseFailure(cause, errorMsg, callId) =>
+            obj.writeField("type").writeString("RPCResponseFailure")
+            obj.writeField("cause").writeString(cause)
+            obj.writeField("errorMsg").writeString(errorMsg)
+            obj.writeField("callId").writeString(callId)
+          case RPCResponseException(name, exception, callId) =>
+            obj.writeField("type").writeString("RPCResponseException")
+            val exceptionField = obj.writeField("exception").writeList()
```
ghik: Actually, in `scala-commons` 1.19.3+, I have introduced a requirement that `ObjectOutput` and `ObjectInput` implementations preserve order of fields. Thanks to that, you can use an object here instead of a list, which would probably be a little more natural. That's because one field can now be deserialized based on value of some other field, which is exactly your case.

For example, look at [current implementation of key-value pair serialization](https://github.com/AVSystem/scala-commons/blob/master/commons-shared/src/main/scala/com/avsystem/commons/serialization/GenCodec.scala#L344).

rpc/shared/shared/src/main/scala/io/udash/rpc/serialization/ExceptionsRegistry.scala
```diff
@@ -0,0 +1,41 @@
...
+      codecs(_name(classTag[T])) = codec
+
+    override def get[T](name: String): GenCodec[T] =
+      codecs(name).asInstanceOf[GenCodec[T]]
+
+    override def contains(name: String): Boolean =
+      codecs.contains(name)
+
+    override def name[T](ex: T): String =
+      ex.getClass.getName
```
ghik: What if codec is registered for superclass or trait? For example, it would be convenient to register a single codec for entire sealed hierarchy of exceptions.

rpc/shared/shared/src/main/scala/io/udash/rpc/serialization/ExceptionsRegistry.scala
```diff
@@ -0,0 +1,41 @@
...
+  def name[T](ex: T): String
+}
+
+object ExceptionsRegistry {
+  val Default: ExceptionsRegistry = new DefaultExceptionsRegistry
+
+  private class DefaultExceptionsRegistry extends ExceptionsRegistry {
+    import scala.reflect._
+
+    private val codecs: mutable.Map[String, GenCodec[_]] = mutable.Map.empty
```
ghik: It might be nice to automatically register default codecs for standard library exceptions that are already available in both JVM and JS (e.g. `IllegalArgumentException` etc.)

rpc/shared/shared/src/main/scala/io/udash/rpc/serialization/ExceptionsRegistry.scala
Starzu: Why not? It's a minor change for us, but someone might want to use custom implementation.

rpc/shared/shared/src/main/scala/io/udash/rpc/UdashRPCFramework.scala
Starzu: That's right, I'm going to change this later.

rpc/shared/shared/src/main/scala/io/udash/rpc/UdashRPCFramework.scala
```diff
@@ -138,69 +116,79 @@ trait UdashRPCFramework extends GetterRPCFramework with ProcedureRPCFramework wi
...
+            obj.writeField("cause").writeString(cause)
+            obj.writeField("errorMsg").writeString(errorMsg)
+            obj.writeField("callId").writeString(callId)
+          case RPCResponseException(name, exception, callId) =>
+            obj.writeField("type").writeString("RPCResponseException")
+            val exceptionField = obj.writeField("exception").writeObject()
+            exceptionField.writeField("type").writeString(name)
+            exceptionsRegistry.get(name).write(exceptionField.writeField("data"), exception)
+            val stack = exceptionField.writeField("stacktrace").writeList()
+            exception.getStackTrace.foreach { element =>
```
ghik: Stacktrace may be absent.

rpc/shared/shared/src/main/scala/io/udash/rpc/serialization/ExceptionCodecRegistry.scala
```diff
@@ -0,0 +1,43 @@
...
+  def contains(name: String): Boolean
+  /** Returns name used for class T. */
+  def name[T <: Throwable](ex: T): String
+}
+
+private[serialization] abstract class ClassNameBasedECR extends ExceptionCodecRegistry {
+  import scala.reflect._
+
+  protected val codecs: mutable.Map[String, GenCodec[_]] = mutable.Map.empty
+  protected val exceptionWriter = (output: Output, ex: Throwable) => output.writeString(if (ex.getMessage != null) ex.getMessage else "")
```
Starzu: It was a temporary change... ;)

rpc/shared/jvm/src/main/scala/io/udash/rpc/serialization/DefaultExceptionCodecRegistry.scala
```diff
@@ -0,0 +1,19 @@
...
+  override def name[T <: Throwable](ex: T): String = {
+    def find(cls: Class[_]): Option[String] = {
+      if (cls == null) None
+      else if (codecs.contains(cls.getName)) Some(cls.getName)
+      else {
+        import com.avsystem.commons._
+        cls.getInterfaces.iterator
+          .flatMap(find)
+          .filter(_.nonEmpty)
+          .nextOpt.toOption
```
ghik: You can simply use `Opt` instead of `Option` in this implementation

Starzu: Right

rpc/shared/shared/src/main/scala/io/udash/rpc/UdashRPCFramework.scala
```diff
@@ -138,69 +116,85 @@ trait UdashRPCFramework extends GetterRPCFramework with ProcedureRPCFramework wi
...
+            RPCResponseSuccess(response.asInstanceOf[RawValue], callId)
+          case "RPCResponseFailure" =>
+            val cause = obj.nextField().assertField("cause").readString()
+            val errorMsg = obj.nextField().assertField("errorMsg").readString()
+            val callId = obj.nextField().assertField("callId").readString()
+            RPCResponseFailure(cause, errorMsg, callId)
+          case "RPCResponseException" =>
+            val wrapper = obj.nextField().assertField("exception").readObject()
+            val exceptionName = wrapper.nextField().assertField("type").readString()
+            val exception = exceptionsRegistry.get[Throwable](exceptionName).read(wrapper.nextField().assertField("data"))
```
ghik: Actually, if we serialize stack trace, we should do it for the whole cause chain... especially considering the fact that the last cause and its stack trace is usually the most interesting.
Possibly we should also serialize all the causes themselves - we can't assume excepton's codec is doing that. Your codecs for standard library exceptions aren't doing that (note: cause can be set on exception using `initCause`)

Starzu: Exceptions serialization is implemented, because it could be useful to match on RPC call's failure type, so I think initial cause is not critical here. Remember that the full stack trace is logged on the server side anyway. 

I'm going to merge this PR now. We can extend serialization mechanism in the future.

ghik: I'd say: either don't serialize stack traces at all or serialize them all. Current version is quite confusingly-incompletish.

Starzu: Right, I'm going revert stack trace serialization for now.

PR #94, opened 2017-03-31
made View.renderChild take an Option[View]

mvh77:
would this be enough do you think? You'll probably have to change the documentation somewhere.

### DIFF
--- a/core/frontend/src/main/scala/io/udash/core/Definitions.scala
+++ b/core/frontend/src/main/scala/io/udash/core/Definitions.scala
@@ -51,11 +51,11 @@ trait View {
     * Will be invoked by [[io.udash.routing.RoutingEngine]] in order to render the child view inside
     * the parent view. <br/><br/>
     *
-    * <b>This method can receive null as "view" argument, then previous child view should be removed.</b>
+    * <b>This method can receive None as "view" argument, then previous child view should be removed.</b>
     *
     * @param view view which origins from child
     */
-  def renderChild(view: View): Unit
+  def renderChild(view: Option[View]): Unit
 
   /**
     * Implementation of this method should return DOM representation of view.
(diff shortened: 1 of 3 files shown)


ddworak:
This is quite a breaking interface change for the users. Maybe we could do something less intrusive?

One thing that comes to my mind is an `OptArg`. See: https://github.com/AVSystem/scala-commons/blob/master/commons-shared/src/main/scala/com/avsystem/commons/misc/OptArg.scala#L28

mvh77:
Ok the tests don't seem to run for me, I get the following error. Anyway, unless there's a quick fix I'm gonna drop this for now. I suggest you get rid of all nulls before 1.0 though, very unScala-like to have to handle nulls.

`[error] (core-frontend/test:loadedTestFrameworks) org.openqa.selenium.WebDriverException: Failed to connect to binary FirefoxBinary(/usr/bin/firefox) on port 7055; process output follows: 
[error] �ення до Firefox.","creator":"Canonical Ltd.","homepageURL":null},{"locales":["zh-CN"],"name":"Ubuntu Modifications","description":"Ubuntu 火狐扩展包.","creator":"Canonical Ltd.","homepageURL":null},{"locales":["zh-TW"],"name":"Ubuntu Modifications","description":"Ubuntu Firefox 擴充包。","creator":"Canonical Ltd.","homepageURL":null}],"targetApplications":[{"id":"{ec8030f7-c20a-464f-9b0e-13a3a9e97384}","minVersion":"9.0","maxVersion":"37.0a1"}],"targetPlatforms":[],"multiprocessCompatible":false,"signedState":2,"seen":true,"dependencies":[],"hasEmbeddedWebExtension":false,"mpcOptedOut":false}
`

ddworak:
The error is related to Selenium WebDriver for Firefox, you need to have it configured properly to run the shared & frontend tests. They are quite flaky on Travis though for the very same reason, so don't be too afraid about errors there.

PR #95, opened 2017-04-05
Charts raw wrapper

Starzu:


Starzu:
Udash Guide examples: https://github.com/UdashFramework/udash-guide/commit/e38bc0284152eef042904dc54e0e1b36d801b736

charts/frontend/src/main/scala/io/udash/wrappers/highcharts/config/tooltip/Tooltip.scala
```diff
@@ -0,0 +1,405 @@
...
+  val xDateFormat: js.UndefOr[String] = js.undefined
+}
+
+object Tooltip {
+  import scala.scalajs.js.JSConverters._
+
+  @js.native
+  trait FormatterData extends js.Object {
+    val percentage: Double = js.native
+    val point: Point = js.native
```
ddworak: I'm afraid one of {`point`, `points`} is undefined depending on the plurality, just Highcharts things...

ddworak:
I think I've seen enough of it in action to accept, could you only verify the unverified (`TODO not verified`)?

Starzu:
@ddworak Done and squashed.

PR #99, opened 2017-05-18
Implicits improvements

ddworak:
* natively support `com.avsystem.commons.misc.Opt`
* consider immutable collections of immutable values valid `ImmutableValue`s


core/macros/src/main/scala/io/udash/macros/PropertyMacros.scala
```diff
@@ -132,10 +139,11 @@ class PropertyMacros(val c: blackbox.Context) {
 
     val isSeq = isImmutableSeq(valueType)
     lazy val isOption = isImmutableOption(valueType)
+    lazy val isOpt = isImmutableOpt(valueType)
     lazy val isSealedHierarchy = isImmutableSealedHierarchy(valueType)
     lazy val isImmutableCC = isImmutableCaseClass(valueType)
 
-    if (isSeq || isOption || isSealedHierarchy || isImmutableCC) {
+    if (isSeq || isOption || isOpt || isSealedHierarchy || isImmutableCC) {
```
Starzu: Add this condition to the error message below.

PR #105, opened 2017-06-12
Backend CSS rendering

### LINKED ISSUE #103: Prepare utils for backend CSS rendering with type-safe frontend usage.
Problem: ScalaCSS generates a lot of JS and makes initialization very slow

Solution:
* it might use DSL from ScalaCSS 
* stylesheets should be defined in cross-compiled module
  * only class names should be compiled into JS file
  * backend should render styles to file or to `String` (and pass them through RPC interface)

Starzu:
Closes #103


css/backend/src/main/scala/io/udash/css/CssFileRenderer.scala
```diff
@@ -0,0 +1,40 @@
...
+import scalacss.internal.Renderer
+
+/**
+  * Renders provided styles into files. Creates separate file for each stylesheet.
+  * Keeps styles order from provided `Seq`.
+  * @param dirPath Target directory for rendered files.
+  * @param styles Sequence of stylesheets.
+  * @param createMain If true, creates `main.css` file. It imports all other stylesheets.
+  */
+class CssFileRenderer(dirPath: String, styles: Seq[_ <: CssBase], createMain: Boolean) {
```
ddworak: You don't need to use an upper-bound for `styles`, `Seq[CssBase]` should be enough.

css/backend/src/main/scala/io/udash/css/CssFileRenderer.scala
```diff
@@ -0,0 +1,40 @@
...
+  def render()(implicit renderer: Renderer[String]): Unit = {
+    val dir = new File(dirPath)
+    dir.mkdirs()
+
+    val mainFile: Option[File] = if (createMain) Some(new File(s"${dir.getAbsolutePath}/main.css")) else None
+    mainFile.foreach(_.createNewFile())
+
+    val mainWriter = mainFile.map(f => new PrintWriter(f, "UTF-8"))
```
ddworak: If you don't care to call the lambda parameter sth else then `f`, just go `.map(new PrintWriter(_, "UTF-8"))`

css/backend/src/main/scala/io/udash/css/CssFileRenderer.scala
```diff
@@ -0,0 +1,40 @@
...
+
+    styles.foreach { style =>
+      val name = style.getClass.getName
+      val f = new File(s"${dir.getAbsolutePath}/$name.css")
+      f.createNewFile()
+      new PrintWriter(f, "UTF-8") {
```
ddworak: I did not ever think of using this syntax for scoping calls to an instance in an anonymous constructor, nice.
Maybe you should do the same with the `new File`s in this class?

css/backend/src/main/scala/io/udash/css/CssFileRenderer.scala
```diff
@@ -0,0 +1,40 @@
...
+        write(style.render)
+        flush()
+        close()
+      }
+
+      mainWriter.foreach(_.append(s"""@import "$name.css";\n"""))
+    }
+
+    mainWriter.foreach(_.flush())
+    mainWriter.foreach(_.close())
```
ddworak: Use single foreach

css/frontend/src/main/scala/io/udash/css/CssView.scala
```diff
@@ -0,0 +1,80 @@
...
+import io.udash.ReadableProperty
+import io.udash.properties.PropertyCreator
+import org.scalajs.dom
+import org.scalajs.dom.Element
+
+import scalatags.JsDom.all.Modifier
+import scalatags.text.Builder
+
+/** Contains integration of CSS structures with Scalatags. */
+trait CssView {
```
ddworak: Extract implicit classes to companion object and provide a conversion here. This will allow you to use value classes.

css/frontend/src/main/scala/io/udash/css/CssView.scala
```diff
@@ -0,0 +1,80 @@
...
+
+    def styleIf(property: ReadableProperty[Boolean]): Modifier =
+      property.reactiveApply(
+        (elem, value) =>
+          if (value) addTo(elem)
+          else removeFrom(elem)
+      )
+
+    def styleIf(condition: Boolean): Modifier =
+      condition match {
```
ddworak: Pattern matching on booleans is not simplified to an if/jump, at least in Scala.

build.sbt
```diff
@@ -168,9 +169,39 @@ lazy val `i18n-frontend` = project.in(file("i18n/frontend"))
...
   )
 
+lazy val `css-macros` = project.in(file("css/macros"))
+  .settings(commonSettings: _*)
+  .settings(
+    libraryDependencies ++= Seq("org.scala-lang" % "scala-reflect" % scalaVersion.value),
+    libraryDependencies ++= cssMacroDeps.value
+  )
+
+lazy val `css-shared` = crossProject.crossType(CrossType.Pure).in(file("css/shared"))
```
ddworak: We might not need `udash-core` dependencies and provide a nice scalacss flavour for current users. This could be a gateway drug to Udash :D

css/shared/src/main/scala/io/udash/css/CssBase.scala
```diff
@@ -0,0 +1,136 @@
...
+
+  private val elementsBuffer = mutable.ArrayBuffer.empty[CssStyle]
+  val dsl: Dsl = new Dsl(elementsBuffer)
+
+  def elements: Seq[CssStyle] =
+    elementsBuffer.toVector
+
+  def render(implicit renderer: Renderer[String]): String = {
+    val sb = new mutable.StringBuilder()
+    elementsBuffer.foreach {
```
ddworak: `elementsBuffer.iterator.map(...).mkString()` would be cleaner.

css/shared/src/main/scala/io/udash/css/CssStyle.scala
```diff
@@ -0,0 +1,12 @@
+package io.udash.css
+
+import scalacss.internal.{FontFace, StyleS}
+
+/** Representation of stylesheet elements. In JS it's always `CssStyleName`. */
+sealed trait CssStyle {
+  val className: String
```
ddworak: Does it have to be a stable identifier? Abstract members are error-prone.

Starzu: I don't want to suggest it can change.

PR #108, opened 2017-06-30
View.renderChild with Option argument

Starzu:
It's time to make some breaking changes. 

Alternative: #109

### DIFF
--- a/core/frontend/src/main/scala/io/udash/core/Definitions.scala
+++ b/core/frontend/src/main/scala/io/udash/core/Definitions.scala
@@ -51,11 +51,11 @@ trait View {
     * Will be invoked by [[io.udash.routing.RoutingEngine]] in order to render the child view inside
     * the parent view. <br/><br/>
     *
-    * <b>This method can receive null as "view" argument, then previous child view should be removed.</b>
+    * <b>This method can receive `None` as "view" argument, then previous child view should be removed.</b>
     *
     * @param view view which origins from child
     */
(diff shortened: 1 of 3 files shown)


core/frontend/src/main/scala/io/udash/core/Definitions.scala
```diff
@@ -68,8 +68,8 @@ trait View {
 
 /** A [[io.udash.core.View]] which does not have any child view. */
 trait FinalView extends View {
-  override def renderChild(view: View): Unit =
-    if (view != null) throw View.UnexpectedChildView
+  override def renderChild(view: Option[View]): Unit =
```
ddworak: Couldn't it override with a more specific argument type `None`?

core/frontend/src/main/scala/io/udash/core/Definitions.scala
```diff
@@ -68,8 +68,8 @@ trait View {
 
 /** A [[io.udash.core.View]] which does not have any child view. */
 trait FinalView extends View {
```
ddworak: Since we're already making major changes, it might be cleaner to just separate the final and non-final views as equal inheritors of View instead of making them a special case handled with an exception. This would move `renderChild` to a meaningful trait for non-final views.

Starzu:
@mkej @ddworak - take a look at #109 - it's an alternative for this PR with more backwards-incompatible changes.

Starzu:
Closing in favor of #109

PR #109, opened 2017-07-10
Views and States hierarchy refactoring

Starzu:
Alternative to #108 

This PR contains more backward incompatible changes:
* `View.renderChild` with `Option[View]` argument (as in #108)
* `ContainerView` and `FinalView` subtypes of `View`
* `ViewPresenter` renamed to `ViewFactory`
* `ContainerViewFactory` and `FinalViewFactory` types with related `ContainerState` and `FinalState` (subtypes of `State`)
* Parent state type changed to `Option[ContainerState]`


core/frontend/src/main/scala/io/udash/routing/RoutingEngine.scala
```diff
@@ -79,39 +80,39 @@ class RoutingEngine[S <: State : ClassTag : ImmutableValue](routingRegistry: Rou
...
   /** @return Property reflecting current routing state */
-  def currentStateProperty: ReadableProperty[S] = currentStateProp.transform(identity)
+  def currentStateProperty: ReadableProperty[HierarchyRoot] = currentStateProp.transform(identity)
 
   @tailrec
-  private def getStatePath(forState: S, acc: List[S] = Nil): List[S] = forState match {
-    case state if state == rootState => rootState :: acc
-    case state: S => getStatePath(state.parentState.asInstanceOf[S], state :: acc)
+  private def getStatePath(forState: Option[HierarchyRoot], acc: List[HierarchyRoot] = Nil): List[HierarchyRoot] = forState match {
+    case Some(state) if state == rootState => rootState :: acc
```
mkej: I think this branch is not necessary if `rootState.parentState` is always meant to return `None`.

core/frontend/src/main/scala/io/udash/Application.scala
```diff
@@ -9,18 +9,18 @@ import scala.reflect.ClassTag
...
+  * @param routingRegistry     [[io.udash.core.RoutingRegistry]] implementation, which will be used to match [[io.udash.core.Url]] to [[io.udash.core.State]]
+  * @param viewFactoryRegistry [[io.udash.core.ViewFactoryRegistry]] implementation, which will be used to match [[io.udash.core.State]] into [[io.udash.core.ViewFactory]]
+  * @param rootState           The instance of [[io.udash.core.State]] which will treated as main state.
+  * @tparam HierarchyRoot Should be a sealed trait which extends [[io.udash.core.State]].
   */
-class Application[S <: State : ClassTag : ImmutableValue](routingRegistry: RoutingRegistry[S],
-                                                          viewPresenterRegistry: ViewPresenterRegistry[S],
-                                                          rootState: S,
-                                                          urlChangeProvider: UrlChangeProvider = WindowUrlChangeProvider) {
+class Application[HierarchyRoot <: State[HierarchyRoot] : ClassTag : ImmutableValue](routingRegistry: RoutingRegistry[HierarchyRoot],
```
ddworak: This class will be heavily extended by library users and rightfully so. Make some of the methods, which guarantee correct behaviour of core udash features final.

core/frontend/src/main/scala/io/udash/core/Definitions.scala
```diff
@@ -31,74 +31,90 @@ trait Presenter[S <: State] {
...
+
+  /**
+    * Will be invoked by [[io.udash.routing.RoutingEngine]] in order to render the child view inside
+    * the parent view. <br/><br/>
+    *
+    * <b>This method can receive `None` as "view" argument, then previous child view should be removed.</b>
+    *
+    * @param view view which origins from child
+    */
+  def renderChild(view: Option[View]): Unit = {
```
ddworak: Do we want the users to override this? If not, let's make it final and update the doc to state the behaviour, not wish for it.

core/frontend/src/main/scala/io/udash/core/Definitions.scala
```diff
@@ -31,74 +31,90 @@ trait Presenter[S <: State] {
...
-/**
-  * The class which should be used to present the state for [[io.udash.routing.RoutingEngine]].
-  */
-trait State {
-  def parentState: State
+/** A [[io.udash.core.View]] which does not have any child view. */
+trait FinalView extends View
+
+/** The class which should be used to present the state for [[io.udash.routing.RoutingEngine]]. */
+trait State[HierarchyRoot <: State[HierarchyRoot]] {
```
ddworak: Maybe we should use a type member instead? See: https://github.com/ghik/opinionated-scala/wiki/Generics-and-type-members#type-member-solution 
It would be easier to migrate existing code, we could get nice benefits from expressing some of the relations introduced by this PR with dependent types and there would be less underscores messing with generics inference.

core/frontend/src/main/scala/io/udash/view/ViewRenderer.scala
```diff
@@ -15,19 +13,27 @@ private[udash] class ViewRenderer(rootElement: => Element) {
   private lazy val endpoint = rootElement
   private val views = mutable.ArrayBuffer[View]()
 
-  private def mergeViews(path: List[View]): View = {
+  private def renderChild(parent: View, child: Option[View]): Unit =
+    parent match {
+      case p: ContainerView =>
+        p.renderChild(child)
+      case _ =>
```
ddworak: It should be possible to use a more complicated data structure than a list/buffer to avoid this case and traverse a statically checked view tree.

core/frontend/src/main/scala/io/udash/core/Definitions.scala
Starzu: It might be useful to override it, for example when you want to animate view changes.

core/frontend/src/main/scala/io/udash/view/ViewRenderer.scala
Starzu: It's not a trivial change, I think we can leave it as a future work.

Starzu:
@ddworak @mkej take a look at review improvements

core/frontend/src/main/scala/io/udash/core/Definitions.scala
```diff
@@ -94,19 +96,20 @@ trait ContainerView extends View {
 trait FinalView extends View
 
 /** The class which should be used to present the state for [[io.udash.routing.RoutingEngine]]. */
-trait State[HierarchyRoot <: State[HierarchyRoot]] {
-  def parentState: Option[ContainerState[HierarchyRoot] with HierarchyRoot]
+trait State {
+  type HierarchyRoot <: State { type HierarchyRoot = State.this.HierarchyRoot }
```
ddworak: Can you write a test for what should and should not compile with this requirement? ScalaTest has a matcher for that.

Starzu: I don't know why, but Scalatest sometimes thinks that broken code is compiling... Anyway take a look at `StateTest` class.

PR #111, opened 2017-07-13
Properties improvements

Starzu:
This PR contains:
* better docs for Properties
* less restrictive Property model checks:
  * allows to create `ModelProperty[T]` with `T` containing implemented defs and vals (both for trait and case classes)
  * ModelProperty treats abstract vals in trait as subproperties
* improved compilation error messages

### DIFF
--- a/core/frontend/src/test/scala/manual/PropertyErrorManualTest.scala
+++ b/core/frontend/src/test/scala/manual/PropertyErrorManualTest.scala
@@ -12,15 +12,19 @@ object PropertyErrorManualTest {
 
   /**
     * Expected error:
+    *
     * `manual.PropertyErrorManualTest.ClassicClass` should meet requirements for one of:
     *   * model value - it has to be an immutable value (io.udash.properties.ImmutableValue[manual.PropertyErrorManualTest.ClassicClass] has to exist), model part or model seq
-    *   * model part - it has to be a trait with abstract methods returning valid model values or a simple case class (contains only vals with valid model values in the primary constructor)
+    *   * model part - it has to be a trait with abstract methods returning valid model values or an immutable case class
     *   * model seq - it has to be Seq[T] where T is a valid model value
     *
     * Try to call one of these methods in your code to get more details:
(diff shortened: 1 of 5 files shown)


core/frontend/src/test/scala/manual/PropertyErrorManualTest.scala
```diff
@@ -206,4 +211,45 @@ object PropertyErrorManualTest {
 //    ImmutableValue.isImmutable[CCB]
 //    ImmutableValue.isImmutable[CCD]
 //  }
+
```
mkej: Added lines below this one should be deleted - no need to keep commented-out test cases.

core/macros/src/main/scala/io/udash/macros/PropertyMacros.scala
```diff
@@ -101,40 +122,68 @@ class PropertyMacros(val c: blackbox.Context) extends MacroCommons {
...
 
   private def checkModel(tree: c.Tree): Boolean =
     c.typecheck(tree, silent = true) != EmptyTree
 
-  private def isModelPart(tpe: Type): Boolean = checkModel(q"""implicitly[$ModelPartCls[$tpe]]""")
+  private def checkModelWithCache(tree: c.Tree, tpe: Type, cache: mutable.Set[Any]): Boolean =
+    if (cache.contains(tpe)) true
+    else {
+      cache += tpe
+      val r = c.typecheck(tree, silent = true) != EmptyTree
```
mkej: Can additional cache lookups be made during `c.typecheck`? If not, wouldn't it be better to just do `if (r) cache += tpe` later on?

core/frontend/src/test/scala/manual/PropertyErrorManualTest.scala
Starzu: In this file all test cases are commented-out, because they don't compile and I use them to manually test compilation errors from properties.

core/macros/src/main/scala/io/udash/macros/PropertyMacros.scala
Starzu: These type checks are recursive.

PR #116, opened 2017-08-09
Authentication utils

### LINKED ISSUE #23: Support for authorization
### TODO
- permissions management system
- frontend bindings for permissions
- utils for RPC endpoints (access verification)

Starzu:
Additionally: 
* StaticViewFactory is no longer abstract
* Views are no longer rendered if presenter thrown an exception in `handleState` method
* Application handles errors in routing

Closes #23

### DIFF
--- a/auth/frontend/src/main/scala/io/udash/auth/AuthPresenter.scala
+++ b/auth/frontend/src/main/scala/io/udash/auth/AuthPresenter.scala
@@ -1,0 +1,19 @@
+package io.udash.auth
+
+import io.udash._
+
+/**
+  * Presenter which check user access in `handleState` method.
+  *
+  * @param permission PermissionCombinator verified against provided `userCtx`.
+  * @param requireAuthenticated If `true`, the presenter requires `userCtx` to don't be Unauthenticated subclass.
+  */
+abstract class AuthPresenter[S <: State](permission: PermissionCombinator, requireAuthenticated: Boolean = false)
+                                        (implicit userCtx: UserCtx)
+  extends Presenter[S] with AuthRequires {
+
+  override def handleState(state: S): Unit = {
+    require(permission, requireAuthenticated)
+  }
+
+}
--- a/auth/shared/src/main/scala/io/udash/auth/AuthRequires.scala
+++ b/auth/shared/src/main/scala/io/udash/auth/AuthRequires.scala
@@ -1,0 +1,15 @@
+package io.udash.auth
+
+trait AuthRequires {
+  /** Checks if user context has required permissions. */
+  def require(permission: PermissionCombinator, requireAuthenticated: Boolean = false)(implicit userCtx: UserCtx): Unit = {
+    if (requireAuthenticated && !userCtx.isAuthenticated) throw new UnauthenticatedException()
+    if (!permission.check(userCtx)) throw new UnauthorizedException()
+  }
+
+  /** Checks if user is authenticated. */
+  def requireAuthenticated()(implicit userCtx: UserCtx): Unit =
+    require(PermissionCombinator.AllowAll, requireAuthenticated = true)
+}
+
+object AuthRequires extends AuthRequires
--- a/auth/shared/src/main/scala/io/udash/auth/Permission.scala
+++ b/auth/shared/src/main/scala/io/udash/auth/Permission.scala
@@ -1,0 +1,32 @@
+package io.udash.auth
+
+import scala.language.implicitConversions
+
+/**
+  * Base class for permissions used with PermissionControl. Permissions are compared by ID.
+  */
+trait Permission {
+  def id: PermissionId
+
+  override def equals(other: Any): Boolean = other match {
+    case that: Permission => id == that.id
+    case _ => false
+  }
+
+  override def hashCode(): Int =
+    id.hashCode()
+
+  override def toString =
+    s"Permission(${id.value}"
+}
+
+object Permission {
+  /** Single permission as a combinator resolved implicitly. */
+  implicit class Single(val permission: Permission) extends AnyVal with PermissionCombinator {
+    override def check(ctx: UserCtx): Boolean =
+      ctx.has(permission)
+
+    override def toString: String =
+      permission.toString
+  }
+}
--- a/auth/shared/src/main/scala/io/udash/auth/Permission.scala
+++ b/auth/shared/src/main/scala/io/udash/auth/Permission.scala
@@ -22,7 +22,7 @@ trait Permission {
 
 object Permission {
   /** Single permission as a combinator resolved implicitly. */
-  implicit class Single(val permission: Permission) extends AnyVal with PermissionCombinator {
+  implicit class Single(private val permission: Permission) extends AnyVal with PermissionCombinator {
     override def check(ctx: UserCtx): Boolean =
       ctx.has(permission)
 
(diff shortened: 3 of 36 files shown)


auth/frontend/src/main/scala/io/udash/auth/AuthPresenter.scala
```diff
@@ -0,0 +1,19 @@
...
+  * Presenter which check user access in `handleState` method.
+  *
+  * @param permission PermissionCombinator verified against provided `userCtx`.
+  * @param requireAuthenticated If `true`, the presenter requires `userCtx` to don't be [[io.udash.auth.UserCtx.Unauthenticated]] subclass.
+  */
+abstract class AuthPresenter[S <: State](permission: PermissionCombinator, requireAuthenticated: Boolean = false)
+                                        (implicit userCtx: UserCtx) extends Presenter[S] {
+
+  override def handleState(state: S): Unit = {
+    if (requireAuthenticated && userCtx.isInstanceOf[UserCtx.Unauthenticated]) throw new UnauthenticatedException()
```
ddworak: If we design user authentication through different context classes but we don't intend to benefit from pattern matching, maybe we should just have one UserCtx class able to answer that?

auth/shared/src/main/scala/io/udash/auth/AuthRequires.scala
```diff
@@ -0,0 +1,15 @@
+package io.udash.auth
+
+trait AuthRequires {
+  /** Checks if user context has required permissions. */
+  def require(permission: PermissionCombinator, requireAuthenticated: Boolean = false)(implicit userCtx: UserCtx): Unit = {
+    if (requireAuthenticated && userCtx.isInstanceOf[UserCtx.Unauthenticated]) throw new UnauthenticatedException()
```
ddworak: This is the 4th implementation of this branching that I've seen so far and this one is actually reusable. Can we only have one?

auth/shared/src/main/scala/io/udash/auth/Permission.scala
```diff
@@ -0,0 +1,34 @@
+package io.udash.auth
+
+import scala.language.implicitConversions
+
+/**
+  * Base class for permissions used with PermissionControl.
+  * Any subclasses have to be defined in shared project as case classes/objects.
```
ddworak: This can be partially enforced by a self type annotation for Product.

auth/shared/src/main/scala/io/udash/auth/Permission.scala
```diff
@@ -0,0 +1,34 @@
...
+  override def hashCode(): Int =
+    id.hashCode()
+
+  override def toString =
+    s"Permission(${id.value}"
+}
+
+object Permission {
+  /** Single permission as a combinator resolved implicitly. */
+  implicit class Single(permission: Permission) extends PermissionCombinator {
```
ddworak: Probably possible to make PermissionCombinator a trait and this a value class.

auth/shared/src/main/scala/io/udash/auth/PermissionId.scala
```diff
@@ -0,0 +1,6 @@
+package io.udash.auth
+
+import com.avsystem.commons.serialization.HasGenCodec
+
+case class PermissionId(value: String) extends AnyVal
```
ddworak: `@transparent`?

auth/shared/src/main/scala/io/udash/auth/PermissionCombinator.scala
```diff
@@ -0,0 +1,60 @@
...
+
+    override def toString: String =
+      s"($this || $other)"
+  }
+}
+
+object PermissionCombinator {
+  /** PermissionCombinator always granting permission. */
+  val AllowAll: PermissionCombinator = new PermissionCombinator {
+    override def check(ctx: UserCtx): Boolean = true
```
ddworak: Does it allow unathenticated users to enter as well? Needs a test and comment.

auth/shared/src/main/scala/io/udash/auth/Permission.scala
Starzu: I'm not gonna lie: I copied this trait from our internal code and I don't have any idea why subclasses have to case class/object. ;)

ddworak: I do, I wrote it :D That was back when GenCodec was the bad guy and we used the automatic mode, so a lot of bad stuff happened when they weren't.

Starzu: I'm going to remove this comment, because there is no such requirement from `Permission` perspective.

auth/shared/src/main/scala/io/udash/auth/Permission.scala
```diff
@@ -24,7 +22,7 @@ trait Permission {
 
 object Permission {
   /** Single permission as a combinator resolved implicitly. */
-  implicit class Single(permission: Permission) extends PermissionCombinator {
+  implicit class Single(val permission: Permission) extends AnyVal with PermissionCombinator {
```
ddworak: private val?

core/frontend/src/main/scala/io/udash/Application.scala
```diff
@@ -39,9 +43,20 @@ class Application[HierarchyRoot <: GState[HierarchyRoot] : ClassTag : ImmutableV
       .recover { case ex: Throwable => handleRoutingFailure(ex) }
   }
 
-  def handleRoutingFailure(ex: Throwable): Unit = {
+  def registerRoutingFailureListener(listener: RoutingFailureListener): Registration = {
```
ddworak: Document and test this public method. The users need to know the ordering and execution guarantees. What happens to other listeners if one of them throws an exception?

rpc/frontend/src/main/scala/io/udash/rpc/DefaultServerRPC.scala
```diff
@@ -26,7 +26,7 @@ object DefaultServerRPC {
             ServerRPCType : DefaultServerUdashRPCFramework.AsRealRPC]
            (localRpc: ClientRPCType, serverUrl: String = "/atm/",
             exceptionsRegistry: ExceptionCodecRegistry = new DefaultExceptionCodecRegistry,
-            rpcFailureInterceptors: Seq[(Throwable) => Unit]): ServerRPCType = {
+            rpcFailureInterceptors: Seq[PartialFunction[Throwable, Any]] = Seq.empty): ServerRPCType = {
```
ddworak: You are using these `PartialFunction`s with `.foreach`, which calls `.apply` and fails when given `Throwable` type is not handled. I'd expect from this method signature that I can intercept specific exception types. I believe only `.collect` checks `.isDefinedAt` before applying the function.

rpc/frontend/src/main/scala/io/udash/rpc/internals/UsesServerRPC.scala
```diff
@@ -44,24 +46,29 @@ private[rpc] trait UsesServerRPC[ServerRPCType] extends UsesRemoteRPC[ServerRPCT
...
-  def registerCallFailureCallback(callback: Throwable => Unit): Registration = {
+  def registerCallFailureCallback(callback: RPCFailureListener): Registration = {
     exceptionCallbacks += callback
     new Registration {
       override def cancel(): Unit = exceptionCallbacks -= callback
     }
   }
 
+  private def handleException(ex: Throwable): Unit =
+    exceptionCallbacks.foreach { pf =>
```
ddworak: This is `.collect`.

Starzu: This is collection of PFs and this `.foreach` applies all of them to provided exception. `.collect` applies provided PF to elements from collection.

Anyway I'll refactor callbacks handing, because it's reusable piece of code.

PR #117, opened 2017-08-10
Properties utils

Starzu:
* Property.set() with force parameter
* Property.listenOnce
* Property.clearListeners
* Listener registration restart


core/frontend/src/main/scala/io/udash/properties/seq/DirectSeqPropertyImpl.scala
```diff
@@ -29,8 +29,10 @@ class DirectSeqPropertyImpl[A](val parent: ReadableProperty[_], override val id:
     valueChanged()
   }
 
-  override def set(t: Seq[A]): Unit =
-    replace(0, properties.size, t: _*)
+  override def set(t: Seq[A], force: Boolean = false): Unit =
+    if (force || t == null || properties.size != t.size || t != get) {
```
ddworak: The third check is kind of evaluated again in the last one, isn't it? IMHO premature optimization.

core/frontend/src/main/scala/io/udash/properties/seq/TransformedSeqProperty.scala
```diff
@@ -30,6 +30,9 @@ class TransformedReadableSeqProperty[A, B, +ElemType <: ReadableProperty[B], Ori
   override def listen(l: (Seq[B]) => Any): Registration =
     origin.listen((seq: Seq[A]) => l(seq.map(transformer)))
 
+  override def listenOnce(l: (Seq[B]) => Any): Registration =
```
ddworak: The arguments here and above need better names, they're meaningful functions.

core/frontend/src/main/scala/io/udash/properties/single/Property.scala
```diff
@@ -69,7 +69,19 @@ trait ReadableProperty[A] {
   /** Registers listener which will be called on value change. */
   def listen(l: A => Any): Registration = {
     listeners += l
-    new PropertyRegistration(listeners, l)
+    new SetRegistration(listeners, l)
+  }
+
+  /** Registers listener which will be called on the next value change. This listener will be fired only once. */
+  def listenOnce(l: A => Any): Registration = {
+    val wrapper: A => Any = new Function1[A, Any] {
```
ddworak: Does this require the full anonymous class syntax?

Starzu: I need `this` reference here

PR #118, opened 2017-08-11
DOM bindings - handling nested bindings clean up

Starzu:
## Problem

Sometimes you want to create property binding inside another binding builder. For example:
```scala
val p = Property("A")
val p2 = Property("a")
produce(p) { v =>
  div(v, bind(p2)).render
}
```
When you change `p` value, the builder creates new element with new binding inside, but unfortunately the old one is still working. There is no way to kill or reuse the old binding, so it will be working as long as you keep reference `p2`.

## Solution

The new variations of bindings (`*WithNested`) providing interceptor which prepares nested binding to be removed when it's no longer needed.  For example:
```scala
val p = Property("A")
val p2 = Property("a")
produceWithNested(p) { (v, nested) =>
  div(v, nested(bind(p2))).render
}
```

### DIFF
--- a/core/frontend/src/main/scala/io/udash/bindings/modifiers/SeqPropertyModifier.scala
+++ b/core/frontend/src/main/scala/io/udash/bindings/modifiers/SeqPropertyModifier.scala
@@ -1,16 +1,44 @@
 package io.udash.bindings.modifiers
 
-import io.udash.bindings._
+import com.avsystem.commons.SharedExtensions._
+import io.udash.bindings.Bindings._
 import io.udash.properties._
 import io.udash.properties.seq.{Patch, ReadableSeqProperty}
 import io.udash.properties.single.ReadableProperty
-import org.scalajs.dom
 import org.scalajs.dom._
 
-import scalatags.generic._
+import scala.scalajs.js
 
-private[bindings] class SeqPropertyModifier[T, E <: ReadableProperty[T]]
-  (property: ReadableSeqProperty[T, E], builder: E => Seq[Element]) extends Modifier[dom.Element] with Bindings {
+private[bindings]
+class SeqPropertyModifier[T, E <: ReadableProperty[T]](property: ReadableSeqProperty[T, E],
+                                                       builder: (E, Binding => Binding) => Seq[Element])
+  extends Binding {
+
+  def this(property: ReadableSeqProperty[T, E], builder: E => Seq[Element]) = {
+    this(property, (d, _) => builder(d))
+  }
+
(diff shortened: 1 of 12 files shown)


core/frontend/src/main/scala/io/udash/bindings/modifiers/SeqPropertyModifier.scala
```diff
@@ -1,16 +1,44 @@
...
+private[bindings]
+class SeqPropertyModifier[T, E <: ReadableProperty[T]](property: ReadableSeqProperty[T, E],
+                                                       builder: (E, Binding => Binding) => Seq[Element])
+  extends Binding {
+
+  def this(property: ReadableSeqProperty[T, E], builder: E => Seq[Element]) = {
+    this(property, (d, _) => builder(d))
+  }
+
+  protected val nestedBindingsByProperty: js.Dictionary[js.Array[Binding]] = js.Dictionary.empty
```
ddworak: Any reason to use a `.js.Dictionary` (and earlier an `js.Array`) instead of Scala collections?

core/frontend/src/main/scala/io/udash/bindings/modifiers/ValueModifier.scala
```diff
@@ -1,35 +1,45 @@
...
+  protected def builder: ((T, Binding => Binding) => Seq[dom.Element])
+  protected def checkNull: Boolean
+  protected def listen(callback: T => Unit): Registration
+
+  protected def replace(root: Element)(oldElements: Seq[Element], newElements: Seq[Element]): Unit =
+    root.replaceChildren(oldElements, newElements)
 
   override def applyTo(t: dom.Element): Unit = {
-    var elements: Seq[Element] = null
+    var elements: Seq[Element] = Seq.empty
```
ddworak: I'd rather see this as an explicitly mutable collection but with an immutable reference.

ddworak:
We should consider a way of warning the user against these nested bindings.

core/frontend/src/main/scala/io/udash/bindings/modifiers/SeqPropertyModifier.scala
Starzu: It's probably premature optimization, but native collections are faster.

core/frontend/src/main/scala/io/udash/bindings/modifiers/ValueModifier.scala
Starzu: It is always replaced with new sequence of elements. I think it's easier to understand than `clear` and `++=` combination.

PR #120, opened 2017-09-08
New modifiers

Starzu:
* showIfElse
* repeatWithIndex

### DIFF
--- a/core/frontend/src/main/scala/io/udash/bindings/modifiers/SeqPropertyWithIndexModifier.scala
+++ b/core/frontend/src/main/scala/io/udash/bindings/modifiers/SeqPropertyWithIndexModifier.scala
@@ -1,0 +1,50 @@
+package io.udash.bindings.modifiers
+
+import com.avsystem.commons.SharedExtensions._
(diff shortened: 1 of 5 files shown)


core/frontend/src/main/scala/io/udash/bindings/modifiers/SeqPropertyWithIndexModifier.scala
```diff
@@ -0,0 +1,44 @@
...
+import scala.scalajs.js
+
+import scala.concurrent.ExecutionContext.Implicits.global
+
+private[bindings]
+class SeqPropertyWithIndexModifier[T, E <: ReadableProperty[T]](override val property: ReadableSeqProperty[T, E],
+                                                                builder: (E, ReadableProperty[Int], Binding => Binding) => Seq[Element])
+  extends SeqPropertyModifierUtils[T, E] {
+
+  private val indexes: mutable.HashMap[E, Property[Int]] = mutable.HashMap.empty
```
mkej: Possible memory leak.

PR #122, opened 2017-09-13
Bindings with customizable DOM manipulators

Starzu:
Additional changes:
* bindings use `Node` instead of `Element`
* modification of listeners set inside listener works fine now

### DIFF
--- a/core/frontend/src/main/scala/io/udash/bindings/modifiers/DOMManipulator.scala
+++ b/core/frontend/src/main/scala/io/udash/bindings/modifiers/DOMManipulator.scala
@@ -4,8 +4,7 @@ import org.scalajs.dom.{Element, Node}
 
 private[bindings]
 trait DOMManipulator {
-  type ReplaceMethod = (Node, Seq[Node], Seq[Node]) => Boolean
-  type InsertMethod = (Node, Node, Seq[Node]) => Boolean
+  import DOMManipulator._
 
   /**
     * Provides custom child elements replace method. This method takes
@@ -19,7 +18,7 @@ trait DOMManipulator {
     * root element, ref node and new children. It should return `true`,
     * if it does not insert elements in DOM.
     */
-  def customElementsInsert: InsertMethod = DOMManipulator.defaultElementInsert
+  def customElementsInsert: InsertMethod = defaultElementInsert
 
   protected def replace(root: Node)(oldElements: Seq[Node], newElements: Seq[Node]): Unit =
     if (customElementsReplace(root, oldElements, newElements)) {
@@ -34,6 +33,9 @@ trait DOMManipulator {
 
 private[bindings]
 object DOMManipulator {
-  val defaultElementReplace: DOMManipulator#ReplaceMethod = (_, _, _) => true
-  val defaultElementInsert: DOMManipulator#InsertMethod = (_, _, _) => true
+  type ReplaceMethod = (Node, Seq[Node], Seq[Node]) => Boolean
+  type InsertMethod = (Node, Node, Seq[Node]) => Boolean
+
+  val defaultElementReplace: ReplaceMethod = (_, _, _) => true
+  val defaultElementInsert: InsertMethod = (_, _, _) => true
 }
--- a/core/frontend/src/main/scala/io/udash/bindings/modifiers/DOMManipulator.scala
+++ b/core/frontend/src/main/scala/io/udash/bindings/modifiers/DOMManipulator.scala
@@ -1,0 +1,39 @@
+package io.udash.bindings.modifiers
+
+import org.scalajs.dom.{Element, Node}
+
+private[bindings]
+trait DOMManipulator {
+  type ReplaceMethod = (Node, Seq[Node], Seq[Node]) => Boolean
+  type InsertMethod = (Node, Node, Seq[Node]) => Boolean
+
+  /**
+    * Provides custom child elements replace method. This method takes
+    * root element, old children and new children. It should return `true`,
+    * if it does not replace elements in DOM.
+    */
+  def customElementsReplace: ReplaceMethod
+
+  /**
+    * Provides custom child elements insert method. This method takes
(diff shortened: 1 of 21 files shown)


core/frontend/src/main/scala/io/udash/bindings/modifiers/DOMManipulator.scala
```diff
@@ -0,0 +1,39 @@
+package io.udash.bindings.modifiers
+
+import org.scalajs.dom.{Element, Node}
+
+private[bindings]
+trait DOMManipulator {
+  type ReplaceMethod = (Node, Seq[Node], Seq[Node]) => Boolean
+  type InsertMethod = (Node, Node, Seq[Node]) => Boolean
```
mkej: Those two should be moved to companion object.

core/frontend/src/main/scala/io/udash/properties/seq/SeqProperty.scala
```diff
@@ -134,8 +134,11 @@ trait ReadableSeqProperty[A, +ElemType <: ReadableProperty[A]] extends ReadableP
   def nonEmpty: Boolean =
     elemProperties.nonEmpty
 
-  protected def fireElementsListeners[ItemType <: ReadableProperty[A]](patch: Patch[ItemType], structureListeners: Iterable[(Patch[ItemType]) => Any]): Unit =
-    CallbackSequencer.queue(s"${this.id.toString}:fireElementsListeners:${patch.hashCode()}", () => structureListeners.foreach(_.apply(patch)))
+  protected def fireElementsListeners[ItemType <: ReadableProperty[A]](patch: Patch[ItemType], structureListeners: Iterable[(Patch[ItemType]) => Any]): Unit = {
+    val cpy = mutable.Set.empty[(Patch[ItemType]) => Any]
```
mkej:
```scala
val cpy = structureListeners.toSet
```

core/frontend/src/main/scala/io/udash/properties/single/Property.scala
```diff
@@ -151,7 +151,9 @@ trait ReadableProperty[A] {
   protected[properties] def fireValueListeners(): Unit = {
     CallbackSequencer.queue(s"${this.id.toString}:fireValueListeners", () => {
       val t = get
-      listeners.foreach(_.apply(t))
+      val cpy = mutable.Set.empty[A => Any]
```
mkej:
```scala
val cpy = listeners.toSet
```

core/frontend/src/main/scala/io/udash/bindings/modifiers/DOMManipulator.scala
```diff
@@ -0,0 +1,39 @@
+package io.udash.bindings.modifiers
+
+import org.scalajs.dom.{Element, Node}
+
+private[bindings]
+trait DOMManipulator {
```
ddworak: What's this line break for?

core/frontend/src/main/scala/io/udash/bindings/modifiers/DOMManipulator.scala
ddworak: I cannot see a reason for using type members here. The types are not overridden anywhere, I'm not sure they even should be and if these contracts were just a part of the method signatures in this trait, the arguments would have meaningful names.

core/frontend/src/main/scala/io/udash/bindings/modifiers/package.scala
```diff
@@ -1,10 +1,10 @@
 package io.udash.bindings
 
-import org.scalajs.dom.Element
+import org.scalajs.dom.Node
 
 package object modifiers {
-  implicit class ElementExts(val el: Element) extends AnyVal {
-    def replaceChildren(oldChildren: Seq[Element], newChildren: Seq[Element]): Unit = {
+  implicit class ElementExts(val el: Node) extends AnyVal {
```
ddworak: private val

core/frontend/src/main/scala/io/udash/bindings/modifiers/DOMManipulator.scala
Starzu: Consistency with the other classes in this package.

core/frontend/src/main/scala/io/udash/bindings/modifiers/DOMManipulator.scala
```diff
@@ -0,0 +1,41 @@
...
+  import DOMManipulator._
+
+  /**
+    * Provides custom child elements replace method. This method takes
+    * root element, old children and new children. It should return `true`,
+    * if it does not replace elements in DOM.
```
ddworak: The result requirements should be explained, so the user understands what that boolean triggers.

core/frontend/src/main/scala/io/udash/bindings/modifiers/DOMManipulator.scala
```diff
@@ -0,0 +1,41 @@
...
+    */
+  def customElementsReplace: ReplaceMethod
+
+  /**
+    * Provides custom child elements insert method. This method takes
+    * root element, ref node and new children. It should return `true`,
+    * if it does not insert elements in DOM.
```
ddworak: Same here.

core/frontend/src/main/scala/io/udash/bindings/modifiers/DOMManipulator.scala
```diff
@@ -0,0 +1,41 @@
...
+      newElements.foreach(root.insertBefore(_, before))
+    }
+}
+
+private[bindings]
+object DOMManipulator {
+  type ReplaceMethod = (Node, Seq[Node], Seq[Node]) => Boolean
+  type InsertMethod = (Node, Node, Seq[Node]) => Boolean
+
+  val defaultElementReplace: ReplaceMethod = (_, _, _) => true
```
ddworak: Make these final and use upper-camel case for constants.

PR #125, opened 2017-10-10
produce() calls the callback even when some properties are null

leonardehrenfried:
This demonstrates the issue displayed in https://scalafiddle.io/sf/uPKEaxc/0

When you have a model property that itself contains sub-properties that are case classes, these are null on an empty model. `produce` does not check this before passing it to its callback.


Starzu:
I have prepared more tests in #127. Thanks for reporting the problem.

PR #130, opened 2017-10-20
Icons - fixed styleIf icons switching

Starzu:
This new test fails with previous Icons implementation


ddworak:
Why didn't they work in the first place?

Starzu:
@ddworak All icons share `fa`/`glyphicon` style and it might be removed with `styleIf` modifier. Take a look at this new test case - it fails with the old implementation.

PR #135, opened 2017-10-28
Add protocol parameter to DefaultServerREST and DefaultRESTConnector

### LINKED ISSUE #132: DefaultServerREST is missing an option to use HTTPS
I'm trying to create a REST client, following the examples in http://guide.udash.io/#/rest/client-server but my backend server is HTTPS only. It seems there is no way to specify HTTPS when using DefaultServerREST.

I'm using udash 0.5.0.

esamson:
This allows client to specify HTTP or HTTPS for REST calls. It
accommodates including the final ':' from the URL scheme part so that
it's convenient to use with a DOM Location object.

    DefaultServerREST[ServerApi](
      dom.window.location.protocol,
      dom.window.location.hostname,
      dom.window.location.port.toInt,
      "/rest"
    )

Fixes #132

### DIFF
--- a/rest/shared/src/main/scala/io/udash/rest/DefaultRESTConnector.scala
+++ b/rest/shared/src/main/scala/io/udash/rest/DefaultRESTConnector.scala
@@ -3,7 +3,7 @@ package io.udash.rest
 import java.nio.ByteBuffer
 
 import monix.execution.Scheduler.Implicits.global
-import fr.hmil.roshttp.{HttpRequest, Method}
+import fr.hmil.roshttp.{HttpRequest, Method, Protocol}
 import fr.hmil.roshttp.body.BodyPart
 import io.udash.rest.internal.RESTConnector
 import io.udash.rest.internal.RESTConnector.HTTPMethod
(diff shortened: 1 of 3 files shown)


rest/shared/src/main/scala/io/udash/rest/DefaultRESTConnector.scala
```diff
@@ -13,13 +13,21 @@ import scala.concurrent.{ExecutionContext, Future}
 import scala.language.implicitConversions
 
 /** Default implementation of [[io.udash.rest.internal.RESTConnector]] for Udash REST. */
-class DefaultRESTConnector(val host: String, val port: Int, val pathPrefix: String)(implicit val ec: ExecutionContext) extends RESTConnector {
+class DefaultRESTConnector(_protocol: String, val host: String, val port: Int, val pathPrefix: String)(implicit val ec: ExecutionContext) extends RESTConnector {
```
ddworak: Would you mind making the protocol an enum (sealed hierarchy)? I like that you did not leak the roshttp internals :+1:, it's just to: 
* avoid throwing that exception below*
* avoid 2 `String` params next to each other are error-prone
* be able to directly expose and use the `protocol`, without further sanitization

rest/shared/src/main/scala/io/udash/rest/DefaultRESTConnector.scala
```diff
@@ -13,13 +13,21 @@ import scala.concurrent.{ExecutionContext, Future}
...
+
+  private val protocol = _protocol match {
```
ddworak: If you follow my previous suggestion, this mapping will probably be inlined / moved to `send()`.

rest/shared/src/main/scala/io/udash/rest/DefaultRESTConnector.scala
leonardehrenfried: Maybe you can provide a combination of the two and add an apply method to the enum. 
Then you can do something like this:
```scala
Protocol(window.location.protocol).getOrElse(HTTP)
```

Would it make sense to provide a default value in the constructor of DefaultRESTConnector?

ddworak: Works for me if you want that kind of flexibility.
As far as the default value is concerned, if there was one, it would be probably be the one you've shown above, but notice that the code is cross-compiled, so there's no DOM assumed (therefore no `window`). Nevertheless it's not something you're going to write every few hours, so I'd prefer the explicit version.

leonardehrenfried: That kind of flexibility would be useful for my current project where the API and the frontend are hosted on the same server.

So in localhost it's http in production it's https.

I see your point about the default value and I agree that it you probably set it up once. I would not suggest to make the default value the result of the string parsing.

ddworak: Oh and by the way, you can use `NamedEnumCompanion`:
https://github.com/AVSystem/scala-commons/blob/master/commons-shared/src/main/scala/com/avsystem/commons/misc/SealedUtils.scala#L100
to neatly implement that apply from `String` (see `byName`).

Starzu:
Thanks @esamson. I'll close it in favor of #138.

PR #138, opened 2017-11-08
Add ability to specify protocol for rest adapters

### LINKED ISSUE #132: DefaultServerREST is missing an option to use HTTPS

leonardehrenfried:
This incorporates the review comments of #135.

It allows you to write the following code:

```scala
 DefaultServerREST[ServerApi](
  Protocol(dom.window.location.protocol).getOrElse(Protocol.Http),
  dom.window.location.hostname,
  dom.window.location.port.toInt,
  "/rest"
)
```

Fixes #132.


rest/shared/src/main/scala/io/udash/rest/DefaultRESTConnector.scala
```diff
@@ -12,14 +12,33 @@ import monix.reactive.Observable
 import scala.concurrent.{ExecutionContext, Future}
 import scala.language.implicitConversions
 
+sealed trait Protocol
+object Protocol {
+  case object Http extends Protocol
+  case object Https extends Protocol
+
+  def apply(s: String): Option[Protocol] = s match {
```
Starzu: I think it should be: 
```s.toLowerCase() match {```

File: rest/shared/src/main/scala/io/udash/rest/DefaultRESTConnector.scala [hunk 2]
```diff
@@ -12,14 +12,33 @@ import monix.reactive.Observable
...
+    case "http:" | "http" => Some(Http)
+    case "https:" | "https" => Some(Https)
+    case _ => None
+  }
+}
+
 /** Default implementation of [[io.udash.rest.internal.RESTConnector]] for Udash REST. */
-class DefaultRESTConnector(val host: String, val port: Int, val pathPrefix: String)(implicit val ec: ExecutionContext) extends RESTConnector {
+class DefaultRESTConnector(protocol: Protocol, val host: String, val port: Int, val pathPrefix: String)(implicit val ec: ExecutionContext) extends RESTConnector {
```
Starzu: Add `val` to `protocol`, it should be consistent with `host`, `port` and `pathPrefix`.

PR #140, opened 2017-11-09
GenCodec instead of GenCodec.Auto in RPC, scala-commons 1.24.0

ghik:
* RPC frameworks now require `GenCodec` instead of `GenCodec.Auto`
* commons 1.24.0
* migrated `RPCTypeClasses` to JS-DCE-friendly `RPCCompanion`


ghik:
After a day of debugging I finally got this pinned down: https://github.com/scala-js/scala-js/issues/3177

PR #141, opened 2017-11-17
Remove all items from a SeqProperty that satisfy a predicate

leonardehrenfried:
Today I wanted to remove an item for a SeqProperty by id and was looking for a method to do this.

It didn't exist so I added it.

The performance implications are worth discussing: I'm not sure if `zipWithIndex` creates an eager or lazy list and therefore if the list is traversed once or twice.


leonardehrenfried:
I realised that my initial implementation, which would collect the idxs to delete first and then call replace on them, was incorrect because the indexes would change as the list was changed.

The current version has to potentially traverse the list n + 1 times where n is the number if items that match the predicate.

PR #142, opened 2017-11-21
built-in support for styles with "prefix" common class name

ghik:


ghik:
Pardon. I ran `+test` but SBT hang up.

PR #143, opened 2017-11-23
JS size and performance optimizations

Starzu:


rpc/shared/shared/src/test/scala/io/udash/rpc/types.scala
```diff
@@ -11,8 +11,7 @@ object NestedTestCC extends HasGenCodec[NestedTestCC]
 case class DeepNestedTestCC(n: NestedTestCC, l: DeepNestedTestCC)
 object DeepNestedTestCC extends HasGenCodec[DeepNestedTestCC]
 
-case class CompleteItem(
-  unit: Unit, string: String, specialString: String, char: Char, boolean: Boolean, byte: Byte, short: Short, int: Int,
-  long: Long, float: Float, double: Double, binary: Array[Byte], list: List[String], set: Set[String],
-  obj: TestCC, map: Map[String, Int])
+case class CompleteItem(unit: Unit, string: String, specialString: String, char: Char, boolean: Boolean, byte: Byte, short: Short, int: Int,
```
ghik: `AbstractCase` will shave off some JS here.

bootstrap/frontend/src/main/scala/io/udash/bootstrap/carousel/UdashCarousel.scala
```diff
@@ -84,7 +85,7 @@ final class UdashCarousel private(content: ReadableSeqProperty[UdashCarouselSlid
         span(`class` := "sr-only", "Next")
       )
     ).render
-    val jqCarousel = jQ(res).asCarousel()
+    val jqCarousel = jQ(res).asInstanceOf[UdashCarouselJQuery]
```
ddworak: What is the impact of all those inlines? Weren't they already inlined or wouldn't they be with `@inline` hint?

bootstrap/frontend/src/main/scala/io/udash/bootstrap/datepicker/UdashDatePicker.scala
```diff
@@ -274,129 +274,125 @@ object UdashDatePicker {
...
+    val keepInvalid: Boolean = false,
+    val ignoreReadonly: Boolean = false,
+    val allowInputToggle: Boolean = false,
+    val focusOnShow: Boolean = true,
+    val enabledHours: Seq[Int] = Seq.empty,
+    val disabledHours: Seq[Int] = Seq.empty,
+    val viewDate: Boolean = false,
+    val tooltips: DatePickerTooltips = new DatePickerTooltips()
+  ) {
+    def copy(minDate: Option[ju.Date] = minDate, maxDate: Option[ju.Date] = maxDate): DatePickerOptions = {
```
ddworak: This cannot be more error-prone and exposes our internal usage as a part of utility API. Is it really worth the unCasing?

bootstrap/frontend/src/main/scala/io/udash/bootstrap/carousel/UdashCarousel.scala
Starzu: It doesn't change JS size, but I think we don't need this boilerplate to hide `.asInstanceOf`.

bootstrap/frontend/src/main/scala/io/udash/bootstrap/datepicker/UdashDatePicker.scala
Starzu: I'll remove this from public API.

PR #150, opened 2017-12-12
Routing definition API cleanup

ghik:
* Renamed `/:/` to `/`
* Moved all the routing-related stuff to `RoutingRegistry` to limit its visibility as much as possible
* Replaced `BidirectionalMacro` with the one defined in commons


ghik:
It's not only about unslashing, also decolonization!

PR #155, opened 2018-01-04
Preserve CSS rules order from scala code

Starzu:
Requires: https://github.com/japgolly/scalacss/pull/128 or https://github.com/japgolly/scalacss/pull/127

### DIFF
--- a/css/shared/src/main/scala/io/udash/css/CssBase.scala
+++ b/css/shared/src/main/scala/io/udash/css/CssBase.scala
@@ -5,7 +5,7 @@ import scala.collection.mutable
 import scala.language.implicitConversions
 import scalacss.internal.Css.styleA
 import scalacss.internal.DslBase.ToStyle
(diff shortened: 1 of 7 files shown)


css/shared/src/main/scala/io/udash/css/CssBase.scala
```diff
@@ -152,17 +153,21 @@ trait CssBase {
...
-              (Percentage(p): scalacss.internal.KeyframeSelector, StyleA(
-                new ClassName(className),
-                Vector.empty,
-                impl
-              ))
-            }
-          ))(Env.empty)
+          CssEntry.Keyframes(
+            kfs.name,
+            ListMap(kfs.frames.map(s => (s._1, styleA(s._2)(Env.empty))):_*)
```
ddworak: 1. Use pattern matching syntax for untupling. 
2. You can avoid copying the mapped sequence with breakOut and explicit type.
```scala
kfs.frames.map(s => (s._1, styleA(s._2)(Env.empty)))(collection.breakOut): ListMap[KeyframeSelector, StyleStream]
```

PR #156, opened 2018-01-05
Client's connection status callbacks

Starzu:


rpc/backend/src/main/scala/io/udash/rpc/DefaultAtmosphereServiceConfig.scala
```diff
@@ -10,11 +11,16 @@ import scala.util.Try
...
   extends AtmosphereServiceConfig[ServerRPCType] with LazyLogging {
 
-  private val RPCName = "RPC"
-  private val connections = new DefaultAtmosphereResourceSessionFactory
+  protected val RPCName = "RPC"
+
+  private val _connections = new DefaultAtmosphereResourceSessionFactory
+  protected def connections: AtmosphereResourceSessionFactory = _connections
+
+  protected val newConnectionCallbacks = new CallbacksHandler[ClientId]
```
ddworak: Make these `final`.

PR #167, opened 2018-03-21
Jawn 0.11.1

### LINKED ISSUE #166: Move to jawn 11.0
My backend uses http4s for making http calls to a different service, which means I depend on circe, which in turn makes jawn 11.0 a transitive dependence. Udash on the other hand is still on 10.4, so on my project I made a fork and hastily changed the code until it would compile and pass the tests (mostly...)

Still, it would be beneficial to move to jawn 11.0 properly, and it shouldn't be too much work.

The changes was made in JawnFacade.scala which is in a hidden folder
(udash-core/rpc/shared/.jvm/src/main/scala/io/udash/rpc/serialization/jawn/JawnFacade.scala) 
so I'm not sure if it's autogenerated or not...

Attaching my changed file, I don't think it's PR worthy, it also makes some tests fail related to which exception should be thrown.

[JawnFacade.scala.tar.gz](https://github.com/UdashFramework/udash-core/files/1829159/JawnFacade.scala.tar.gz)

Starzu:
Closes #166

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -18,8 +18,8 @@ object Dependencies {
   val atmosphereJSVersion = "2.3.4"
   val atmosphereVersion = "2.4.15"
 
-  val upickleVersion = "0.4.4" // Tests only
-  val jawnParserVersion = "0.10.4"
+  val upickleVersion = "0.5.1" // Tests only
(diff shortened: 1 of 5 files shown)


Starzu:
#170 removes JAWN

PR #169, opened 2018-03-22
Property sync with test, AttrOps binds return Binding

FreexD:


core/shared/src/main/scala/io/udash/properties/single/Property.scala
```diff
@@ -253,5 +253,35 @@ trait Property[A] extends ReadableProperty[A] {
     */
   def transformToSeq[B : PropertyCreator](transformer: A => Seq[B], revert: Seq[B] => A): SeqProperty[B, Property[B]] =
     new SeqPropertyFromSingleValue(this, transformer, revert)
+
+  /**
+    * Bidirectionally synchronizes Property[B] with `this`.
```
Starzu: We should notice that it puts transformed value from `this` to `p` on initialization.

core/shared/src/test/scala/io/udash/properties/PropertyTest.scala
```diff
@@ -811,6 +811,98 @@ class PropertyTest extends UdashSharedTest {
...
+
+      // Target update
+      target.set(1)
+
+      source.get should be(1)
+      target.get should be(1)
+
+      // Restart streaming, source touch
+      registration.restart()
+      source.touch()
```
Starzu: I think `restart` should trigger synchronization like `sync` initial call.

PR #170, opened 2018-03-29
Use JsonStringInputOutput for serialization

### LINKED ISSUE #166: Move to jawn 11.0

ddworak:
Closes #166


rpc/shared/.jvm/src/test/scala/io/udash/rpc/JVMSerializationIntegrationTest.scala
```diff
@@ -1,7 +1,7 @@
 package io.udash.rpc
 
 class JVMSerializationIntegrationTest extends SerializationIntegrationTestBase {
-  "DefaultUdashRPCFramework -> DefaultUdashRPCFramework default serialization" should tests(DefaultServerUdashRPCFramework, DefaultServerUdashRPCFramework)
-  "DefaultUdashRPCFramework -> UPickleUdashRPCFramework default serialization" should tests(DefaultServerUdashRPCFramework, ServerUPickleUdashRPCFramework)
```
Starzu: Could you describe possible compatibility problems? I assume incompatibility is the reason to remove these tests.

ddworak: https://scastie.scala-lang.org/hd61gI4vT3WyAsRHSazG3A

PR #172, opened 2018-04-12
Fixed double-encoding of URL arguments in REST client

### LINKED ISSUE #173: Double Encoding in Udash REST API
Hi (To Whom It May Concern):

I was trying to use Udash for a simple project recently.

I used the Udash REST API to call a backend designed using the Flight PHP framework (http://www.flightphp.com).

The backend URLs were database queries which failed to produce the correct result on Chrome Browser v45. On debugging, I found that the URLs called by the REST API (via URLPart or standard query string parameter) had been double-encoded; the spaces " " were turned into "%2520" instead of "%20".

I thought I should use POST, but the Udash documentation isn't very clear on on how to pass POST parameters to the backend API: Is it via @Query or @Body?

So, I ended up manually de-encoding the entity myself.

Could this bug be patched?

Starzu:
Closes #173

### DIFF
--- a/rest/backend/src/main/scala/io/udash/rest/server/ExposesREST.scala
+++ b/rest/backend/src/main/scala/io/udash/rest/server/ExposesREST.scala
@@ -34,7 +34,10 @@ abstract class ExposesREST[ServerRPCType : UdashRESTFramework#ValidServerREST](l
 
   def handleRestCall(req: HttpServletRequest, httpMethod: Class[_ <: RESTMethod])(implicit ec: ExecutionContext): Future[String] = {
     val invocations = List.newBuilder[RawInvocation]
-    val path: Array[String] = Option(req.getPathInfo).map(_.stripPrefix("/").split("/")).getOrElse(Array.empty[String])
+    val path: Array[String] =
+      Option(req.getRequestURI.stripPrefix(req.getServletPath))
+        .map(_.stripPrefix("/").split("/"))
+        .getOrElse(Array.empty[String])
     lazy val bodyContent = req.getReader.lines().toArray.mkString("\n")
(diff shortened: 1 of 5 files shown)


rest/backend/src/main/scala/io/udash/rest/server/ExposesREST.scala
```diff
@@ -34,7 +34,10 @@ abstract class ExposesREST[ServerRPCType : UdashRESTFramework#ValidServerREST](l
 
   def handleRestCall(req: HttpServletRequest, httpMethod: Class[_ <: RESTMethod])(implicit ec: ExecutionContext): Future[String] = {
     val invocations = List.newBuilder[RawInvocation]
-    val path: Array[String] = Option(req.getPathInfo).map(_.stripPrefix("/").split("/")).getOrElse(Array.empty[String])
+    val path: Array[String] =
+      Option(req.getRequestURI.stripPrefix(req.getServletPath))
```
ddworak: Can `.stripPrefix` return null? I'd be more worried about `getRequestURI`. Also use `Opt` :rocket:

rest/shared/src/main/scala/io/udash/rest/internal/UsesREST.scala
```diff
@@ -114,7 +114,7 @@ abstract class UsesREST[ServerRPCType : UdashRESTFramework#AsRealRPC : UdashREST
     }
 
     connector.send(
-      url = s"/${urlBuilder.result().map(URLEncoder.encode).mkString("/")}",
+      url = s"/${urlBuilder.result().mkString("/")}",
```
ddworak: Can we have a test case which to guard against such bugs?

PR #176, opened 2018-05-09
Immutable properties

Starzu:


core/shared/src/main/scala/io/udash/properties/ImmutableProperty.scala
```diff
@@ -0,0 +1,98 @@
...
+  override def transformToSeq[B: PropertyCreator](transformer: A => Seq[B]): ReadableSeqProperty[B, ReadableProperty[B]] =
+    new ImmutableSeqProperty[B](transformer(value))
+
+  override def streamTo[B](target: Property[B], initUpdate: Boolean)(transformer: A => B): Registration = {
+    if (initUpdate) target.set(transformer(value))
+    ImmutableProperty.noopRegistration
+  }
+}
+
+private[properties] class ImmutableSeqProperty[A](value: Seq[A]) extends ImmutableProperty[Seq[A]](value) with ReadableSeqProperty[A, ImmutableProperty[A]] {
```
ddworak: It's not exactly immutable if there can be a `mutable.Seq` in it.

core/shared/src/main/scala/io/udash/properties/Properties.scala
```diff
@@ -25,6 +25,18 @@ trait Properties {
...
+
+  implicit class Any2Property[A](val value: A) {
+    def toProperty: ReadableProperty[A] = new ImmutableProperty[A](value)
+  }
+
+  implicit class Any2ModelProperty[A : ModelPropertyCreator](val value: A) {
+    def toModelProperty: ReadableModelProperty[A] = new ImmutableProperty[A](value)
+  }
+
+  implicit class Any2SeqProperty[A](val value: Seq[A]) {
```
ddworak: Make these 3 value classes with implicit def (as in `scala-commons` extension methods).

core/shared/src/main/scala/io/udash/properties/seq/ReadableSeqProperty.scala
```diff
@@ -0,0 +1,139 @@
...
+
+import io.udash.properties._
+import io.udash.properties.single.{AbstractReadableProperty, ReadableProperty}
+import io.udash.utils.{Registration, SetRegistration}
+
+import scala.collection.mutable
+import scala.concurrent.Future
+
+/** Read-only interface of SeqProperty[A]. */
+trait ReadableSeqProperty[A, +ElemType <: ReadableProperty[A]] extends ReadableProperty[Seq[A]] {
```
ddworak: Same issue with the `Seq`. I'd suggest banning them everywhere.

core/macros/src/main/scala/io/udash/macros/PropertyMacros.scala
```diff
@@ -305,16 +336,21 @@ class PropertyMacros(val ctx: blackbox.Context) extends AbstractMacroCommons(ctx
 
     def genTree(source: List[(Select, TermName)], targetTree: Tree): Tree = source match {
       case (select, term) :: _ if select.tpe.typeConstructor =:= SeqTpe.typeConstructor =>
-        q"""$targetTree.getSubProperty[${select.tpe.widen}](${term.decodedName.toString})
-           .asInstanceOf[$SeqPropertyCls[${select.tpe.typeArgs.head.widen}, $PropertyCls[${select.tpe.typeArgs.head.widen}] with $CastablePropertyCls[${select.tpe.typeArgs.head.widen}]]]"""
+        q"""$targetTree.getSubSeq[${select.tpe.typeArgs.head.widen}](${q"_.$term"}, ${term.decodedName.toString})"""
+      case (select, term) :: Nil if hasModelPropertyCreator(select.tpe.widen) =>
+        q"""{
+            val tmp = $targetTree
+            tmp.getSubModel[${select.tpe.widen}](${q"_.$term"}, ${term.decodedName.toString}).asInstanceOf[tmp.ModelSubProperty[${select.tpe.widen}]]
```
ddworak: The cast and `ModelSubProperty` type seem to be redundant.

core/shared/src/main/scala/io/udash/properties/ImmutableProperty.scala
Starzu: Mutability of data passed to properties is ignored now. You can put anything into `Property` since we've dropped `ImmutableValue`. To keep behavior of `ImmutableSeqProperty` similar to the standard `SeqProperty` implementation we should copy the provided sequence if it's mutable.

PR #177, opened 2018-05-09
Use sequential property IDs

ddworak:
UUIDs can be slow to generate and we don't necesarily need to use random IDs there (especially from a SecureRandom on the JVM side, which is the case, I believe, for that particular `UUID.randomUUID` implementation). 

On my machine, this proved to be quite beneficial for the immutable property binding benchmark (2x improvement). Please verify, it could be a nice addition to your immutable props PR.


Starzu:
That was my plan for tomorrow. ;)

PR #182, opened 2018-05-15
Expose RPC calls timeout in constructor

Starzu:


Starzu:
There is a test for that: `ServerRPCTest:106`

ddworak:
Please don't merge before I the PR is accepted :roll_eyes: 

This test uses `retrying` to test whether there is a timeout at all, but it does not test whether this exposed timeout is used. I can rewrite `io.udash.rpc.DefaultServerRPC#apply` to completely ignore the argument and pass a constant like:
```scala
lazy val serverRPC: DefaultServerRPC[ServerRPCType] = new DefaultServerRPC[ServerRPCType](serverConnector, 1.second)
```
and the tests still pass.

Starzu:
The test does not use `io.udash.rpc.DefaultServerRPC#apply`, because it has to create a mocked connector. I don't think it makes sens to test argument forwarding. What regards to the test itself, we can measure time from the call to the successful escape from `retrying` and verify if it's similar to the call timeout. What do you think?

ddworak:
That sounds good. It does not make sense to test forwarding, but it might well not just be forwarding in the future, so I just want to ensure this with an integration test (obviously not a very precise one).

PR #185, opened 2018-05-23
Path based URL provider

### LINKED ISSUE #98: Support urls without hash
Hi, many popular client side frameworks now support urls without hash (#). It would be nice to have this feature in udash. Thank you!

Starzu:
Closes #98 
Take a look at Udash Guide migration: https://github.com/UdashFramework/udash-guide/commit/cc54f57cc2128e446e1df1c29f65d0baa97c6fc9

### DIFF
--- a/core/frontend/src/main/scala/io/udash/routing/RoutingRegistry.scala
+++ b/core/frontend/src/main/scala/io/udash/routing/RoutingRegistry.scala
@@ -1,0 +1,42 @@
+package io.udash.routing
+
+import io.udash.core.{State, Url}
+
+/**
+  * The implementation of this trait should be injected to [[io.udash.routing.RoutingEngine]].
+  * It should implement a bidirectional mapping between [[io.udash.core.Url]] and [[io.udash.core.State]].
+  */
+trait RoutingRegistry[HierarchyRoot <: State] {
+  def matchUrl(url: Url): HierarchyRoot
+  def matchState(state: HierarchyRoot): Url
+
+  protected def bidirectional(pf: PartialFunction[String, HierarchyRoot]): (PartialFunction[String, HierarchyRoot], PartialFunction[HierarchyRoot, String]) =
+  macro com.avsystem.commons.macros.misc.BidirectionalMacro.impl[String, HierarchyRoot]
+
+  import RoutingRegistry._
+
+  protected final val / = RoutingRegistry./
+
+  protected implicit def stringRoutingOps(str: String): StringRoutingOps =
+    new StringRoutingOps(str)
+}
+
+object RoutingRegistry {
+  implicit class StringRoutingOps(val left: String) extends AnyVal {
+    def /(right: Any): String = RoutingRegistry./(left, right.toString)
+  }
+
+  object / {
+    def unapply(path: String): Option[(String, String)] = {
+      val strippedPath = path.stripSuffix("/")
+      Some(strippedPath.lastIndexOf("/")).filter(_ >= 0).map { splitIndex =>
+        val left = strippedPath.substring(0, splitIndex)
+        val right = strippedPath.substring(splitIndex + 1, strippedPath.length)
+        (left, right)
+      }
+    }
+
+    def apply(left: String, right: String): String =
+      left + "/" + right
+  }
+}
--- a/core/frontend/src/main/scala/io/udash/routing/RoutingRegistry.scala
+++ b/core/frontend/src/main/scala/io/udash/routing/RoutingRegistry.scala
@@ -22,17 +22,18 @@ trait RoutingRegistry[HierarchyRoot <: State] {
 }
 
 object RoutingRegistry {
-  implicit class StringRoutingOps(val left: String) extends AnyVal {
+  implicit final class StringRoutingOps(private val left: String) extends AnyVal {
     def /(right: Any): String = RoutingRegistry./(left, right.toString)
   }
 
   object / {
     def unapply(path: String): Option[(String, String)] = {
       val strippedPath = path.stripSuffix("/")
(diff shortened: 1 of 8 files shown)


core/frontend/src/main/scala/io/udash/routing/RoutingRegistry.scala
```diff
@@ -0,0 +1,42 @@
...
+
+object RoutingRegistry {
+  implicit class StringRoutingOps(val left: String) extends AnyVal {
+    def /(right: Any): String = RoutingRegistry./(left, right.toString)
+  }
+
+  object / {
+    def unapply(path: String): Option[(String, String)] = {
+      val strippedPath = path.stripSuffix("/")
+      Some(strippedPath.lastIndexOf("/")).filter(_ >= 0).map { splitIndex =>
```
ddworak: `filter` + `map` -> `collect`

core/frontend/src/main/scala/io/udash/routing/RoutingRegistry.scala
```diff
@@ -0,0 +1,42 @@
...
+        val left = strippedPath.substring(0, splitIndex)
+        val right = strippedPath.substring(splitIndex + 1, strippedPath.length)
+        (left, right)
```
ddworak: These three lines can be replaced with:
```scala
strippedPath.splitAt(splitIndex)
```

core/frontend/src/main/scala/io/udash/routing/RoutingRegistry.scala
```diff
@@ -0,0 +1,42 @@
...
+  import RoutingRegistry._
+
+  protected final val / = RoutingRegistry./
+
+  protected implicit def stringRoutingOps(str: String): StringRoutingOps =
+    new StringRoutingOps(str)
+}
+
+object RoutingRegistry {
+  implicit class StringRoutingOps(val left: String) extends AnyVal {
```
ddworak: `final`, `private val`

core/frontend/src/main/scala/io/udash/routing/UrlChangeProvider.scala
```diff
@@ -2,41 +2,100 @@ package io.udash.routing
...
-  private var callbacks: List[Url => Unit] = List()
-
-  override def onFragmentChange(callback: Url => Unit): Unit = {
-    callbacks = callback :: callbacks
-  }
-
-  def currentFragment: Url = Url(document.location.hash.stripPrefix("#"))
+  private val callbacks: js.Array[Url => Unit] = js.Array()
 
   Window.onFragmentChange(() => {
```
ddworak: We only ever use `Window` to store a single callback - maybe we could just add an event listener to `DomWindow` here and encapsulate it better (or use `window.document` and get rid of it)?

core/frontend/src/main/scala/io/udash/routing/UrlChangeProvider.scala
```diff
@@ -2,41 +2,100 @@ package io.udash.routing
...
   import dom.document
 
-  private var callbacks: List[Url => Unit] = List()
-
-  override def onFragmentChange(callback: Url => Unit): Unit = {
-    callbacks = callback :: callbacks
-  }
-
-  def currentFragment: Url = Url(document.location.hash.stripPrefix("#"))
+  private val callbacks: js.Array[Url => Unit] = js.Array()
```
ddworak: I believe there's only a single library-registered callback here and we don't provide means to deregister callbacks here, so maybe we could not expose this low-level mechanism to the users? Client apps should prefer `io.udash.Application#currentStateProperty` anyway.

core/frontend/src/main/scala/io/udash/routing/UrlChangeProvider.scala
```diff
@@ -2,41 +2,100 @@ package io.udash.routing
...
+      val href = target.getAttribute("href")
+      val ignore = event.button != 0 || event.metaKey || event.ctrlKey || event.shiftKey ||
+        event.defaultPrevented || target.hasAttribute("download") ||
+        target.getAttribute("rel") == "external" || href.contains("mailto:") || !isSameOrigin(href)
+
+      if (!ignore) {
+        if (!isSamePath(href)) {
+          val url = Url(href)
+          changeFragment(url)
+          callbacks.foreach(_.apply(url))
```
ddworak: `changeFragment` already does that

core/frontend/src/main/scala/io/udash/routing/UrlChangeProvider.scala
```diff
@@ -2,41 +2,100 @@ package io.udash.routing
...
+  override def currentFragment: Url = Url(document.location.hash.stripPrefix("#"))
+  override def changeFragment(url: Url): Unit = document.location.hash = url.value
+  override def changeUrl(url: String): Unit = document.location.replace(url)
 }
+
+/**
+  * Used for routing based on the URL path.
+  * Don't forget to configure your web server to handle frontend routes. You may find "rewrite rules" mechanism useful.
+  */
+object WindowUrlPathChangeProvider extends UrlChangeProvider {
```
ddworak: We need a Selenium test for both providers (can be a part of guide).

core/frontend/src/main/scala/io/udash/routing/RoutingRegistry.scala
Starzu: Nope, `right` skips the `/`.

core/frontend/src/main/scala/io/udash/routing/UrlChangeProvider.scala
Starzu: I think these objects may be useful for different purposes than Udash routing. Anyway we should provide callback removal API if we want to left this method public.

PR #186, opened 2018-05-29
Input bindings - refactoring

### LINKED ISSUE #161: RadioButtons should support generic Properties
Bi-Directional data binding for RadioButtons (and similar inputs) currently only work for String-Properties. This is an unnecessary restriction that causes a lot of boiler plate. Here is an suggestion for the implementation of a typesafe RadioButtons:

```scala
object TypedRadioButtons {
  def apply[A, B](property: Property[A], options: Seq[(A, B)],
      decorator: Seq[(html.Input, A, B)] => JsDom.TypedTag[html.Element], xs: Modifier*): 
  JsDom.TypedTag[html.Element] = {
    val uuid = java.util.UUID.randomUUID().toString
    decorator(options.map(opt => {
      val in = input(tpe := "radio", name := uuid, onchange :+= ((event: Event) => {
        property.set(opt._1)
      }))(xs: _*).render
      in.checked = property.get == opt._1
      property.listen { _ =>
        in.checked = property.get == opt._1
      }
      (in, opt._1, opt._2)
    }))
  }

}
```

It takes any property of type A, a Seq, that contains Tuples associating the property values with some additional information of type B. That information can be used to generate the labels, so it can be text or some html elements.

Starzu:
* improved cleanup process
* always two arguments lists: behavior and apperance
* more generic API of selectors (closes #161)

### DIFF
--- a/core/frontend/src/main/scala/io/udash/bindings/inputs/CheckButtons.scala
+++ b/core/frontend/src/main/scala/io/udash/bindings/inputs/CheckButtons.scala
@@ -10,22 +10,23 @@ import scalatags.JsDom.all._
 
 /** Checkboxes for finite options with many elements selection. Bound to SeqProperty. */
 object CheckButtons {
-  def inputsOnlyDecorator[T]: Seq[(JSInput, T)] => Seq[Node] = RadioButtons.inputsOnlyDecorator
+  def inputsOnlyDecorator[T]: Seq[(JSInput, T)] => Seq[Node] =
+    RadioButtons.inputsOnlyDecorator
   def spanWithLabelDecorator[T](labelFactory: T => Modifier): Seq[(JSInput, T)] => Seq[Node] =
     RadioButtons.spanWithLabelDecorator(labelFactory)
   def divWithLabelDecorator[T](labelFactory: T => Modifier): Seq[(JSInput, T)] => Seq[Node] =
     RadioButtons.divWithLabelDecorator(labelFactory)
 
   /**
-    * @param selectedItems SeqProperty which gonna be bound to checkboxes
+    * @param selectedItems SeqProperty which is going to be bound to checkboxes
     * @param options Seq of available options, one checkbox will be created for each option.
     * @param decorator Function creating HTML element from checkboxes Seq.
-    * @param xs Modifiers to apply on each generated checkbox.
+    * @param inputModifiers Modifiers to apply on each generated checkbox.
     * @return HTML element created by decorator.
     */
   def apply[T : PropertyCreator](
     selectedItems: SeqProperty[T, _ <: ReadableProperty[T]], options: ReadableProperty[Seq[T]]
-  )(decorator: Seq[(JSInput, T)] => Seq[Node], xs: Modifier*): InputBinding[Div] = {
+  )(decorator: Seq[(JSInput, T)] => Seq[Node], inputModifiers: Modifier*): InputBinding[Div] = {
     new InputBinding[Div] {
       private val buttons = div(
         nestedInterceptor(
@@ -35,7 +36,7 @@ object CheckButtons {
             decorator(
               opts.zipWithIndex.map { case (opt, idx) =>
                 val in = input(
-                  xs, tpe := "checkbox", value := idx.toString,
+                  inputModifiers, tpe := "checkbox", value := idx.toString,
                   nested((checked := "checked").attrIf(selectedItems.transform(_.contains(opt))))
                 ).render
 
@@ -56,7 +57,7 @@ object CheckButtons {
   }
 
   /**
-    * @param property SeqProperty which gonna be bound to checkboxes
+    * @param property SeqProperty which is going to be bound to checkboxes
     * @param options Seq of available options, one checkbox will be created for each option.
     * @param decorator Function creating HTML element from checkboxes Seq.
     * @param xs Modifiers to apply on each generated checkbox.
--- a/core/frontend/src/main/scala/io/udash/bindings/inputs/CheckButtons.scala
+++ b/core/frontend/src/main/scala/io/udash/bindings/inputs/CheckButtons.scala
@@ -2,7 +2,9 @@ package io.udash.bindings.inputs
 
 import io.udash.properties.seq.SeqProperty
 import io.udash.properties.single.ReadableProperty
-import org.scalajs.dom.{html, _}
+import org.scalajs.dom.{Element, Event}
+import org.scalajs.dom.html.{Input => JSInput}
+
 import scalatags.JsDom
 import scalatags.JsDom.all._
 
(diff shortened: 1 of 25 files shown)


core/frontend/src/test/scala/io/udash/bindings/inputs/SelectTest.scala
```diff
@@ -0,0 +1,313 @@
...
+
+  "Select with multiple on" should {
+    "synchronise state with property changes (deprecated)" in {
+      val options = Seq("A", "B", "C", "D", "E")
+      val p = SeqProperty[String]("B")
+
+      val select = (Select(p, options, Select.defaultLabel)(): @silent).render
+
+      select.childElementCount should be(5)
+      select.childNodes(0).asInstanceOf[org.scalajs.dom.html.Option].selected should be(false)
```
ghik: Maybe some aliased import for `Option` instead of repeating FQCN bajillion times?

ghik: Also, can you make it shorter, e.g.

```scala
List(false, true, false, false, false).zipWithIndex.foreach {
  case (value, index) => select.childNodes(index).asInstanceOf[HtmlOption].selected should be(value)
}
```
?

core/frontend/src/main/scala/io/udash/bindings/inputs/CheckButtons.scala
```diff
@@ -0,0 +1,99 @@
...
+/** Checkboxes for finite options with many elements selection. Bound to SeqProperty. */
+object CheckButtons {
+  def inputsOnlyDecorator[T]: Seq[(JSInput, T)] => Seq[Node] = RadioButtons.inputsOnlyDecorator
+  def spanWithLabelDecorator[T](labelFactory: T => Modifier): Seq[(JSInput, T)] => Seq[Node] =
+    RadioButtons.spanWithLabelDecorator(labelFactory)
+  def divWithLabelDecorator[T](labelFactory: T => Modifier): Seq[(JSInput, T)] => Seq[Node] =
+    RadioButtons.divWithLabelDecorator(labelFactory)
+
+  /**
+    * @param selectedItems SeqProperty which gonna be bound to checkboxes
```
ddworak: is going to

core/frontend/src/main/scala/io/udash/bindings/inputs/CheckButtons.scala
```diff
@@ -0,0 +1,99 @@
...
+    * @param options Seq of available options, one checkbox will be created for each option.
+    * @param decorator Function creating HTML element from checkboxes Seq.
+    * @param xs Modifiers to apply on each generated checkbox.
+    * @return HTML element created by decorator.
+    */
+  def apply[T : PropertyCreator](
+    selectedItems: SeqProperty[T, _ <: ReadableProperty[T]], options: ReadableProperty[Seq[T]]
+  )(decorator: Seq[(JSInput, T)] => Seq[Node], xs: Modifier*): InputBinding[Div] = {
```
ddworak: Modifiers usually work nicely as a separate list. Also, there are a few better names than `xs` :rofl:

core/frontend/src/main/scala/io/udash/bindings/inputs/CheckButtons.scala
```diff
@@ -0,0 +1,99 @@
...
+
+  /**
+    * @param selectedItems SeqProperty which gonna be bound to checkboxes
+    * @param options Seq of available options, one checkbox will be created for each option.
+    * @param decorator Function creating HTML element from checkboxes Seq.
+    * @param xs Modifiers to apply on each generated checkbox.
+    * @return HTML element created by decorator.
+    */
+  def apply[T : PropertyCreator](
+    selectedItems: SeqProperty[T, _ <: ReadableProperty[T]], options: ReadableProperty[Seq[T]]
```
ddworak: Is there a reason not to use the `SeqProperty[T]` alias?

core/frontend/src/main/scala/io/udash/bindings/inputs/Input.scala
```diff
@@ -0,0 +1,79 @@
...
+      override def render: JSInput = element
+    }
+
+  /**
+    * @param property Property to bind.
+    * @param debounce Property update timeout after input changes.
+    * @param xs Additional Modifiers, don't use modifiers on value, onchange and onkeyup attributes.
+    * @return HTML textarea with bound Property, applied modifiers and nested options.
+    */
+  @deprecated("Use `apply` returning `InputBinding` instead.", "0.7.0")
```
ddworak: You can probably just remove these methods since this object is `private[bindings]`

core/frontend/src/main/scala/io/udash/bindings/inputs/Input.scala
```diff
@@ -0,0 +1,79 @@
...
+  private class InputModifier(property: Property[String], debounce: Option[Duration])
+    extends TextInputsModifier(property, debounce)  {
+
+    override def elementValue(t: Element): String =
+      t.asInstanceOf[JSInput].value
+
+    override def setElementValue(t: Element, v: String): Unit =
+      t.asInstanceOf[JSInput].value = if (v != null) v else ""
+
+    override def setElementKeyUp(t: Element, callback: KeyboardEvent => Any): Unit =
```
ddworak: Wouldn't `KeyboardEvent => Unit` make more sense?

core/frontend/src/main/scala/io/udash/bindings/inputs/RadioButtons.scala
```diff
@@ -0,0 +1,104 @@
...
+import org.scalajs.dom.{Element, Event, Node}
+import org.scalajs.dom.html.{Div, Input => JSInput}
+import scalatags.JsDom
+import scalatags.JsDom.all._
+
+/**
+  * Radio buttons group for finite options with one element selection.
+  */
+object RadioButtons {
+  def inputsOnlyDecorator[T]: Seq[(JSInput, T)] => Seq[Node] = _.map(_._1)
```
ddworak: Can we default to ISeq everywhere? These are all new methods.

Also: use pattern matching instead of `._1`

core/frontend/src/main/scala/io/udash/bindings/inputs/RadioButtons.scala
```diff
@@ -0,0 +1,104 @@
...
+                (in, opt)
+              }
+            )
+          }
+        )
+      ).render
+
+      override def render: Div = buttons
+    }
+  }
```
ddworak: This is almost 1:1 the code from `CheckButtons`. Maybe we could introduce an abstraction?

core/frontend/src/main/scala/io/udash/bindings/inputs/RadioButtons.scala
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+    * @param property Property which gonna be bound to radio buttons group.
```
ddworak: We all gonna go to hell :laughing:

core/frontend/src/main/scala/io/udash/bindings/inputs/RadioButtons.scala
```diff
@@ -0,0 +1,104 @@
...
+        val element = t.asInstanceOf[JSInput]
+
+        updateInput(element)
+        property.listen(_ => updateInput(element))
+        element.onchange = (_: Event) => {
+          property.set(element.value)
+        }
+      }
+    }
+  }
```
ddworak: This also seems like a copy from CheckButtons

core/frontend/src/main/scala/io/udash/bindings/inputs/Input.scala
Starzu: This private abstract class is extended by public objects `TextInput`, `PasswordInput` and `NumberInput`.

core/frontend/src/main/scala/io/udash/bindings/inputs/RadioButtons.scala
Starzu: Not really, if you meant these three decorators by "everywhere". `Seq` in `SeqProperty` is the root of this problem.

core/frontend/src/main/scala/io/udash/bindings/inputs/CheckButtons.scala
Starzu: I want too keep all these applies similar. First list describes behavior, second apperance of element.

core/frontend/src/main/scala/io/udash/bindings/inputs/CheckButtons.scala
Starzu: The alias requires Property as subproperties type.

PR #191, opened 2018-06-06
Input bindings fixes

Starzu:
Fixes to #186


core/frontend/src/main/scala/io/udash/bindings/inputs/RadioButtons.scala
```diff
@@ -32,7 +32,7 @@ object RadioButtons {
     selectedItem: Property[T], options: ReadableProperty[Seq[T]]
```
ddworak: Is there a reason to use `ReadableProperty[Seq]` here and a `SeqProperty` in the `CheckButtons`? We could probably avoid rerendering everything if these were both SeqProperties.

Starzu: `CheckButtons` uses `ReadableProperty[Seq]` too. I think that `produce` is good enough here, but we can change the argument type to `ReadableSeqProperty` to make replacement with `repeat` backward compatible.

Issue #189, opened 2018-06-04
Seq[Property[T]] combine


FreexD:
Add a method to combine Seq[Property[T]] into ReadableSeqProperty[T].

PR #192, opened 2018-06-06
Seq[Property] combine extension method

FreexD:
Closes https://github.com/UdashFramework/udash-core/issues/189

### DIFF
--- a/core/shared/src/main/scala/io/udash/properties/PropertySeqCombinedReadableSeqProperty.scala
+++ b/core/shared/src/main/scala/io/udash/properties/PropertySeqCombinedReadableSeqProperty.scala
@@ -1,32 +1,77 @@
 package io.udash.properties
 
 import com.avsystem.commons.ISeq
-import io.udash.properties.Properties.{Patch, Property, ReadableProperty}
+import io.udash.properties.Properties.{Patch, ReadableProperty}
 import io.udash.properties.seq.AbstractReadableSeqProperty
 import io.udash.utils.Registration
 
-private[properties] final class PropertySeqCombinedReadableSeqProperty[A](value: ISeq[Property[A]])
+private[properties]
+class PropertySeqCombinedReadableSeqProperty[A](value: ISeq[ReadableProperty[A]])
   extends AbstractReadableSeqProperty[A, ReadableProperty[A]] {
 
   override val id: PropertyId = PropertyCreator.newID()
   override protected[properties] val parent: ReadableProperty[_] = null
 
-  private val children = CrossCollections.createArray[ReadableProperty[A]]
+  private val children = value.map(_.readable)
+  private var originListenerRegistration: Registration = _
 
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/io/udash/properties/PropertySeqCombinedReadableSeqProperty.scala
```diff
@@ -0,0 +1,37 @@
...
+private[properties] final class PropertySeqCombinedReadableSeqProperty[A](value: ISeq[Property[A]])
+  extends AbstractReadableSeqProperty[A, ReadableProperty[A]] {
+
+  override val id: PropertyId = PropertyCreator.newID()
+  override protected[properties] val parent: ReadableProperty[_] = null
+
+  private val children = CrossCollections.createArray[ReadableProperty[A]]
+
+  value.foreach(property => {
+    children.+=(property.transform(value => value))
```
Starzu: This type of transform is not a good way to ensure that no one can access mutable API with pattern matching. I know that there is no better way now, but it's time to provide it. I think method `_.readonly` returning a minimal wrapper around the property will be much better. (Ping @ddworak)

core/shared/src/main/scala/io/udash/properties/PropertySeqCombinedReadableSeqProperty.scala
```diff
@@ -0,0 +1,37 @@
...
+    property.listen(_ => valueChanged())
```
Starzu: This implementation does not allow GC to clean up this property, when it's no longer needed, but the origin `property` is still in use. Take a look at `transform` implementation from `ReadableProperty`.

core/shared/src/main/scala/io/udash/properties/PropertySeqCombinedReadableSeqProperty.scala
```diff
@@ -0,0 +1,37 @@
+package io.udash.properties
+
+import com.avsystem.commons.ISeq
+import io.udash.properties.Properties.{Patch, Property, ReadableProperty}
+import io.udash.properties.seq.AbstractReadableSeqProperty
+import io.udash.utils.Registration
+
+private[properties] final class PropertySeqCombinedReadableSeqProperty[A](value: ISeq[Property[A]])
```
Starzu: I think that `ISeq[ReadableProperty[A]]` is enough.

core/shared/src/main/scala/io/udash/properties/PropertySeqCombinedReadableSeqProperty.scala
ddworak: True, one can always pattern match on the property at runtime and find out it has some nice mutable features.

Starzu: #196

Issue #188, opened 2018-06-04
Style bindings


FreexD:
Add Style.bind method - for cases when we need to bind property value (or transformed property value) into a inlined style like `width`, `backgroundColor` - not to use reactiveApply all the time.

PR #193, opened 2018-06-06
InlineStyleOps for convenient inline styles reactive property binding

FreexD:
Closes https://github.com/UdashFramework/udash-core/issues/188

### DIFF
--- a/core/frontend/src/main/scala/io/udash/bindings/Bindings.scala
+++ b/core/frontend/src/main/scala/io/udash/bindings/Bindings.scala
@@ -7,7 +7,7 @@ import io.udash.properties.ValidationResult
 import io.udash.properties.seq.ReadableSeqProperty
 import io.udash.properties.single.ReadableProperty
 import org.scalajs.dom._
-import org.scalajs.dom.raw._
+import org.scalajs.dom.raw.HTMLElement
 import scalatags.JsDom
 import scalatags.generic.{Attr, AttrPair, AttrValue, Modifier}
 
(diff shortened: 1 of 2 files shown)


core/frontend/src/test/scala/io/udash/bindings/TagsBindingTest.scala
```diff
@@ -1904,4 +1904,48 @@ class TagsBindingTest extends UdashFrontendTest with Bindings { bindings: Bindin
...
+      testDiv.style.getPropertyValue("background-color") should ===("black")
+      testDiv.style.getPropertyValue("width") should ===("100px")
+      styleProperty.set(null)
+      pixelStyleProperty.set(null)
+      testDiv.style.getPropertyValue("background-color") should ===("")
+      testDiv.style.getPropertyValue("width") should ===("")
+      styleProperty.set("blue")
+      pixelStyleProperty.set("2rem")
+      testDiv.style.getPropertyValue("background-color") should ===("blue")
+      testDiv.style.getPropertyValue("width") should ===("2rem")
```
Starzu: Important part of the `Binding` interface is the `kill` method. We should test if it clears the listeners properly. 
For example:
```scala
val bindStyle = backgroundColor.bind(styleProperty)
val bindPixelStyle = width.bind(pixelStyleProperty)
val testDiv = div(bindStyle, bindPixelStyle).render
...
styleProperty.listenersCount() should be(1)
pixelStyleProperty.listenersCount() should be(1)

bindStyle.kill()
styleProperty.listenersCount() should be(0)
pixelStyleProperty.listenersCount() should be(1)

bindPixelStyle.kill()
styleProperty.listenersCount() should be(0)
pixelStyleProperty.listenersCount() should be(0)
```

core/frontend/src/main/scala/io/udash/bindings/Bindings.scala
```diff
@@ -395,6 +396,36 @@ trait Bindings {
...
+        case (elem: HTMLElement, v) => applyTo(elem, v)
+        case _ =>
+      }
+
+    /**
+      * Synchronises style value with property content by adding it when property is not null and
+      * condition property is 'true' and removing otherwise.
+      */
+    def bindIf(property: ReadableProperty[String], conditionProperty: ReadableProperty[Boolean]): Binding =
+      property.combine(conditionProperty)((_, _)).reactiveApply {
```
Starzu: As you will see in the proposed tests extension this binding does not clear the listeners properly. There are two possible solutions: 
* comment out the listeners count checks and report an issue to fix the `combine` method,
* fix the `combine` method (it should not register listeners in the origin properties as long as there is no listener in the combined property itself - take a look at the implementation of `transform` method).

core/frontend/src/main/scala/io/udash/bindings/Bindings.scala
```diff
@@ -509,6 +540,19 @@ object Bindings extends Bindings {
     }
   }
 
+  /** Abstraction over scalatags styles representation. */
+  trait HasCssName[-T] {
+    def cssName(v: T): String
+  }
+  object HasCssName {
+    implicit val StyleHasCssName: HasCssName[scalatags.generic.Style] = new HasCssName[scalatags.generic.Style] {
```
Starzu: Use local `import scalatags.generic.{Style, PixelStyle}`, please.

core/frontend/src/main/scala/io/udash/bindings/Bindings.scala
Starzu: #195

PR #203, opened 2018-06-28
Routing URL providers aliases in io.udash._

Starzu:


core/frontend/src/main/scala/io/udash/routing/Routing.scala
```diff
@@ -5,5 +5,7 @@ import io.udash.GState
 trait Routing {
   type RoutingEngine[S >: Null <: GState[S]] = io.udash.routing.RoutingEngine[S]
   type UrlChangeProvider = io.udash.routing.UrlChangeProvider
+  val WindowUrlFragmentChangeProvider = io.udash.routing.WindowUrlFragmentChangeProvider
+  val WindowUrlPathChangeProvider = io.udash.routing.WindowUrlPathChangeProvider
```
ddworak: These should probably be `final`

PR #211, opened 2018-07-26
Toggle method for boolean property

Crossy147:
I came up with that idea after seeing dozens of constructs like ```prop.set(!prop.get)```
Yet to be discussed whether it is a good idea to introduce it as Property's method.


core/shared/src/test/scala/io/udash/properties/PropertyTest.scala
```diff
@@ -96,6 +96,22 @@ class PropertyTest extends UdashSharedTest {
...
+      p.get shouldBe false
+      p.toggle()
+      p.get shouldBe true
+
+      val blank = Property.blank[Boolean]
+      blank.toggle()
+      blank.get shouldBe true
+
+      """Property[String]("asd").toggle()""" shouldNot typeCheck
+      """Property[AnyVal](true).toggle()""" shouldNot typeCheck
```
ddworak: Add a check for `java.lang.Boolean`

Starzu:
I agree with @ddworak. Implicit class `PropertyOps` is the way to go.

PR #213, opened 2018-08-01
Use sttp for default HTTP clients in udash-rest

ddworak:


rest/shared/src/main/scala/io/udash/rest/internal/RESTConnector.scala
```diff
@@ -6,25 +6,23 @@ import scala.concurrent.Future
...
 }
 
 object RESTConnector {
-  sealed trait HTTPMethod
-  case object GET extends HTTPMethod
-  case object POST extends HTTPMethod
-  case object PATCH extends HTTPMethod
-  case object PUT extends HTTPMethod
-  case object DELETE extends HTTPMethod
+  sealed trait HttpMethod
```
Starzu: `AbstractValueEnum`?

PR #216, opened 2018-09-04
Callback successful upload

bartpr:
At some cases after upload, developed action may require processing HTTP response (e.g. getting uploaded file id from servlet after successful request) 

Considered alternative to callback: `case class Completed(response: HttpResponse)`  in upload state model and setting listener on it. Imho, keeping response every time upload has been completed is undesired.

### DIFF
--- a/core/frontend/src/main/scala/io/udash/utils/FileUploader.scala
+++ b/core/frontend/src/main/scala/io/udash/utils/FileUploader.scala
@@ -9,6 +9,7 @@ import scala.scalajs.js
 
 class FileUploader(url: Url) {
   import FileUploader._
+  import FileUploader.FileUploadState.HttpResponse
 
   /** Uploads files selected in provided `input`. */
   def upload(input: html.Input): ReadableModelProperty[FileUploadModel] =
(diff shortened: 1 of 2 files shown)


core/frontend/src/main/scala/io/udash/utils/FileUploader.scala
```diff
@@ -18,8 +18,12 @@ class FileUploader(url: Url) {
     )
 
   /** Uploads provided `files` in a field named `fieldName`. */
-  def upload(fieldName: String, files: Seq[File],
-             extraData: Map[js.Any, js.Any] = Map.empty): ReadableModelProperty[FileUploadModel] = {
+  def upload(
+    fieldName: String,
+    files: Seq[File],
+    extraData: Map[js.Any, js.Any] = Map.empty,
+    successfulResponseCallback: Option[HttpResponse => Unit] = None
```
Starzu: I think that we should add `Option[HttpResponse]` to `FileUploadModel` and set this value similary to `state` field instead of providing this callback.

core/frontend/src/main/scala/io/udash/utils/HttpResponse.scala
```diff
@@ -0,0 +1,26 @@
+package io.udash.utils
+
+import com.avsystem.commons.misc.AbstractCase
+import org.scalajs.dom.Document
+import org.scalajs.dom.raw.XMLHttpRequest
+
+sealed trait HttpResponse {
```
Starzu: I think you don't need this sealed trait, just create a single case class. You can also move it to `FileUploader` companion object. 

We should add information about `status` code and provide access to response headers (`getResponseHeader`) from XHR.

PR #217, opened 2018-09-18
Fixed problem with wildcards in generic subproperties

Starzu:
The problem: https://scalafiddle.io/sf/XtFliC9/0

### DIFF
--- a/core/macros/src/main/scala/io/udash/macros/PropertyMacros.scala
+++ b/core/macros/src/main/scala/io/udash/macros/PropertyMacros.scala
@@ -327,26 +327,31 @@ class PropertyMacros(val ctx: blackbox.Context) extends AbstractMacroCommons(ctx
 
     checkIfIsValidPath(modelPath)
 
-    def parsePath(tree: Tree, acc: List[(Select, TermName)] = List()): List[(Select, TermName)] = tree match {
-      case s@Select(next, t@TermName(_)) => parsePath(next, (s, t) :: acc)
-      case _ => acc
+    def parsePath(tree: Tree): List[(Type, TermName)] = {
+      def symbolPath(t: Tree): List[Symbol] = t match {
+        case Select(pre, _) => t.symbol :: symbolPath(pre)
+        case _ => Nil
+      }
+      def mkPath(prefix: Type, syms: List[Symbol]): List[(Type, Symbol)] = syms match {
+        case Nil => Nil
+        case head :: tail => (prefix, head) :: mkPath(head.typeSignatureIn(prefix), tail)
+      }
+      mkPath(weakTypeOf[A], symbolPath(tree).reverse).map {
+        case (prefixTpe, symbol) => (symbol.typeSignatureIn(prefixTpe).resultType, symbol.asTerm.name)
+      }
     }
 
     val parts = parsePath(modelPath)
 
-    def genTree(source: List[(Select, TermName)], targetTree: Tree): Tree = source match {
-      case (select, term) :: _ if select.tpe.typeConstructor =:= SeqTpe.typeConstructor =>
-        val widenTpe = select.tpe.typeArgs.head.widen
-        q"""$targetTree.getSubSeq[$widenTpe](${q"_.$term"}, ${term.decodedName.toString})"""
-      case (select, term) :: Nil if hasModelPropertyCreator(select.tpe.widen) =>
-        val widenTpe = select.tpe.widen
-        q"""$targetTree.getSubModel[$widenTpe](${q"_.$term"}, ${term.decodedName.toString})"""
-      case (select, term) :: tail if hasModelPropertyCreator(select.tpe.widen) =>
-        val widenTpe = select.tpe.widen
-        genTree(tail, q"""$targetTree.getSubModel[$widenTpe](${q"_.$term"}, ${term.decodedName.toString}).asInstanceOf[$ModelPropertyMacroApiCls[$widenTpe]]""")
-      case (select, term) :: _ =>
-        val widenTpe = select.tpe.widen
-        q"""$targetTree.getSubProperty[$widenTpe](${q"_.$term.asInstanceOf[$widenTpe]"}, ${term.decodedName.toString})"""
+    def genTree(source: List[(Type, TermName)], targetTree: Tree): Tree = source match {
+      case (resultType, term) :: _ if resultType.typeConstructor =:= SeqTpe.typeConstructor =>
+        q"""$targetTree.getSubSeq[${resultType.typeArgs.head}](${q"_.$term"}, ${term.decodedName.toString})"""
+      case (resultType, term) :: Nil if hasModelPropertyCreator(resultType) =>
+        q"""$targetTree.getSubModel[$resultType](${q"_.$term"}, ${term.decodedName.toString})"""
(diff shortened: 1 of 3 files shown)


core/macros/src/main/scala/io/udash/macros/PropertyMacros.scala
```diff
@@ -336,16 +336,20 @@ class PropertyMacros(val ctx: blackbox.Context) extends AbstractMacroCommons(ctx
 
     def genTree(source: List[(Select, TermName)], targetTree: Tree): Tree = source match {
       case (select, term) :: _ if select.tpe.typeConstructor =:= SeqTpe.typeConstructor =>
-        q"""$targetTree.getSubSeq[${select.tpe.typeArgs.head.widen}](${q"_.$term"}, ${term.decodedName.toString})"""
+        val widenTpe = select.tpe.typeArgs.head.widen
+        q"""$targetTree.getSubSeq[$widenTpe](${q"_.$term.asInstanceOf[Seq[$widenTpe]]"}, ${term.decodedName.toString})"""
```
ghik: Why are casts needed?

core/macros/src/main/scala/io/udash/macros/PropertyMacros.scala
```diff
@@ -336,16 +336,20 @@ class PropertyMacros(val ctx: blackbox.Context) extends AbstractMacroCommons(ctx
...
+            tmp.getSubModel[$widenTpe](${q"_.$term.asInstanceOf[$widenTpe]"}, ${term.decodedName.toString})
         }"""
       case (select, term) :: tail if hasModelPropertyCreator(select.tpe.widen) =>
-        genTree(tail, q"""$targetTree.getSubModel[${select.tpe.widen}](${q"_.$term"}, ${term.decodedName.toString}).asInstanceOf[$ModelPropertyMacroApiCls[${select.tpe.widen}]]""")
+        val widenTpe = select.tpe.widen
+        genTree(tail, q"""$targetTree.getSubModel[$widenTpe](${q"_.$term.asInstanceOf[$widenTpe]"}, ${term.decodedName.toString}).asInstanceOf[$ModelPropertyMacroApiCls[$widenTpe]]""")
       case (select, term) :: _ =>
-        q"""$targetTree.getSubProperty[${select.tpe.widen}](${q"_.$term"}, ${term.decodedName.toString})"""
+        val widenTpe = select.tpe.widen
+        q"""$targetTree.getSubProperty[$widenTpe](${q"_.$term.asInstanceOf[$widenTpe]"}, ${term.decodedName.toString})"""
```
ddworak: This cast is the only one caught by tests when removed. If other other 3 are also required, we should have it covered there.

Starzu: Right, we don't need the other casts. A type with a wildcard cannot have `ModelPropertyCreator`. It's similar in case of `Seq[_]`.

core/macros/src/main/scala/io/udash/macros/PropertyMacros.scala
```diff
@@ -380,8 +377,13 @@ class PropertyMacros(val ctx: blackbox.Context) extends AbstractMacroCommons(ctx
   def reifyPropertyCreator[A: c.WeakTypeTag]: c.Tree = {
     val tpe = weakTypeOf[A].dealias
 
-    if (!tpe.typeSymbol.isClass) c.abort(c.enclosingPosition, s"Implicit PropertyCreator[$tpe] not found.")
-    else q"new $SinglePropertyCreatorCls[$tpe]"
+    if (!tpe.typeSymbol.isClass) {
+      c.abort(c.enclosingPosition, s"Implicit PropertyCreator[$tpe] not found.")
+    } else if (tpe =:= SeqTpe) {
+      c.abort(c.enclosingPosition,
+        s"Implicit PropertyCreator[Seq[_]] not found. If you use Seq[_] in your model, replace it with Seq[Any]."
```
ghik:
```
scala> typeOf[List[_]] =:= typeOf[List[Any]]
res3: Boolean = false

scala> typeOf[Seq[_]] <:< typeOf[Seq[Any]]
res4: Boolean = true

scala> typeOf[Seq[Any]] <:< typeOf[Seq[_]]
res5: Boolean = true
```

I love Scala

Starzu: The problem here is:
* for `Seq[_]` the compiler does not want to use `PropertyCreatorImplicits.materializeSeq` and takes `PropertyCreatorImplicitsLow.materializeSingle`
* for `Seq[Any]` is uses `PropertyCreatorImplicits.materializeSeq` as expected

core/macros/src/main/scala/io/udash/macros/PropertyMacros.scala
```diff
@@ -336,16 +336,17 @@ class PropertyMacros(val ctx: blackbox.Context) extends AbstractMacroCommons(ctx
...
+        val widenTpe = select.tpe.widen
+        q"""$targetTree.getSubModel[$widenTpe](${q"_.$term"}, ${term.decodedName.toString})"""
       case (select, term) :: tail if hasModelPropertyCreator(select.tpe.widen) =>
-        genTree(tail, q"""$targetTree.getSubModel[${select.tpe.widen}](${q"_.$term"}, ${term.decodedName.toString}).asInstanceOf[$ModelPropertyMacroApiCls[${select.tpe.widen}]]""")
+        val widenTpe = select.tpe.widen
+        genTree(tail, q"""$targetTree.getSubModel[$widenTpe](${q"_.$term"}, ${term.decodedName.toString}).asInstanceOf[$ModelPropertyMacroApiCls[$widenTpe]]""")
       case (select, term) :: _ =>
-        q"""$targetTree.getSubProperty[${select.tpe.widen}](${q"_.$term"}, ${term.decodedName.toString})"""
+        val widenTpe = select.tpe.widen
+        q"""$targetTree.getSubProperty[$widenTpe](${q"_.$term.asInstanceOf[$widenTpe]"}, ${term.decodedName.toString})"""
```
ghik: Ok, I played with this a but trying to understand what _actually_ is going on here and I came to a very professional conclusion that `select.tpe` returns some kind of _bullshit_ type with an unbound wildcard that I have no idea where it comes from.

However, a more "principled" solution than casting might be to obtain less broken type. Instead of computing `List[(Select, TermName)]` you could compute `List[(Type, Symbol)]` where the type is the type of the prefix on which the symbol is selected. The type of selection itself could be computed as `symbol.typeSignatureIn(prefixTpe)`.

ghik: That's a quick proposition on how to compute `List[(Type, Symbol)]`, although I haven't tested it:

```scala
    def parsePath(tree: Tree): List[(Type, Symbol)] = {
      def symbolPath(t: Tree): List[Symbol] = t match {
        case Select(pre, _) => t.symbol :: symbolPath(pre)
        case _ => Nil
      }
      def mkPath(prefix: Type, syms: List[Symbol]): List[(Type, Symbol)] = syms match {
        case Nil => Nil
        case head :: tail => (prefix, head) :: mkPath(head.typeSignatureIn(prefix), tail)
      }
      mkPath(weakTypeOf[A], symbolPath(tree))
    }
```

ghik:
:+1:

PR #220, opened 2018-10-15
Project structure simplified

ghik:
* removed `shared` / `backend` / `frontend` distinction and replaced it with single cross-compiled project (e.g. `core` + `core-js` with shared sources)
* merged all macro projects into one
* various minor project structure improvements


build.sbt
```diff
@@ -24,30 +38,16 @@ inThisBuild(Seq(
...
-// Settings for JS tests run in browser
-val browserCapabilities: Capabilities = {
-  // requires ChromeDriver: https://sites.google.com/a/chromium.org/chromedriver/
-    val options = new ChromeOptions()
-    options.addArguments("--headless", "--disable-gpu")
-    options
-}
-
-val commonSettings = Seq(
   moduleName := "udash-" + moduleName.value,
```
Starzu: Since we're changing the modules structure anyway, I think we could remove `udash-` prefix. What do you think?

ghik: Hmm, jars would be named like `core-x.y.z.jar` which might easily conflict with whatever other library is using the same convention (remember that UMP puts all jars into one folder).

PR #221, opened 2018-10-16
CssStyle companion object with implicit val blank;

essentin:


css/shared/src/main/scala/io/udash/css/CssStyle.scala
```diff
@@ -9,6 +11,11 @@ sealed trait CssStyle {
   def commonPrefixClass: Option[String] = None
   final def classNames: Seq[String] = commonPrefixClass.toList :+ className
 }
+
+object CssStyle {
+  implicit val blank: Blank[CssStyle] = Simple(CssStyleName(""))
```
Starzu: The empty string is not a valid value for `CssStyleName`.

If you try to execute the following code in your browser console:
```
document.getElementsByTagName("div")[0].classList.add("")
```
it throws the error:
```
SyntaxError: An invalid or illegal string was specified
```

I think we should not provide any `Blank[CssStyle]`, but a model should contain `Option[CssStyle]` in cases when the style can be absent. 

Ping @ddworak @FreexD - what do you think about it?

ddworak: Wouldn't it usually use multiple styles anyway? `Seq[CssStyle]` works perfectly then.

Starzu: Is it a problem to wrap a style into `Seq` by yourself?

FreexD: `Opt`, `Option` and `Seq` all have implicit conversions to `Modifier` on evidence ;) you don't need another conversion for this when you could just simply `apply` whichever option you need ;)

PR #225, opened 2018-10-23
New REST, ported from commons

ghik:


rest/src/main/scala/io/udash/rest/RawRest.scala
```diff
@@ -0,0 +1,238 @@
+package io.udash
+package rest
+
+import java.util.concurrent.atomic.AtomicBoolean
+
+import com.avsystem.commons._
+import com.avsystem.commons.meta._
+import com.avsystem.commons.rpc._
+
+sealed abstract class RestMethodCall {
```
Starzu: There's a lot of internal stuff in `io.udash.rest` package. what about hiding it in `internat` package?

rest/src/main/scala/io/udash/rest/openapi/OpenApi.scala
```diff
@@ -0,0 +1,599 @@
...
+  info: Info,
+  paths: Paths,
+  @td servers: List[Server] = Nil,
+  @td components: OptArg[Components] = OptArg.Empty,
+  @td security: List[SecurityRequirement] = Nil,
+  @td tags: List[Tag] = Nil,
+  @td externalDocs: OptArg[ExternalDocumentation] = OptArg.Empty
+)
+object OpenApi extends HasGenCodec[OpenApi] {
+  final val Version = "3.0.1"
```
Starzu: It's `3.0.2` now.

rest/src/main/scala/io/udash/rest/SttpRestClient.scala
```diff
@@ -0,0 +1,67 @@
...
+
+import com.avsystem.commons._
+import com.avsystem.commons.annotation.explicitGenerics
+import com.avsystem.commons.meta.Mapping
+import com.softwaremill.sttp.Uri.QueryFragment.KeyValue
+import com.softwaremill.sttp.Uri.QueryFragmentEncoding
+import com.softwaremill.sttp._
+import io.udash.rest.DefaultSttpBackend.backend
+
+object SttpRestClient {
```
Starzu: Maybe `DefaultRestClient`?

rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
```diff
@@ -0,0 +1,93 @@
...
+}
+
+object RestServlet {
+  final val DefaultHandleTimeout = 30.seconds
+
+  def apply[@explicitGenerics RestApi: RawRest.AsRawRpc : RestMetadata](
+    apiImpl: RestApi, handleTimeout: FiniteDuration = DefaultHandleTimeout
+  ): RestServlet = new RestServlet(RawRest.asHandleRequest[RestApi](apiImpl))
+
+  def readParameters(request: HttpServletRequest): RestParameters = {
```
Starzu: `private`?

rest/src/main/scala/io/udash/rest/SttpRestClient.scala
```diff
@@ -0,0 +1,67 @@
...
+  def apply[@explicitGenerics RestApi: RawRest.AsRealRpc : RestMetadata](baseUri: String): RestApi =
```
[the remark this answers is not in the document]
ghik: The doc is right ;)

rest/src/main/scala/io/udash/rest/openapi/OpenApi.scala
ghik: The code I wrote based on 3.0.1, so we can assume this is the version generated ;)

rest/src/main/scala/io/udash/rest/RawRest.scala
ghik: Introduced `raw` package, but I'd like to leave these things public anyway.

rest/src/main/scala/io/udash/rest/DefaultRestApiCompanion.scala
```diff
@@ -0,0 +1,264 @@
...
+import com.avsystem.commons.rpc.{AsRaw, AsRawReal, AsReal, InvalidRpcCall}
+import com.avsystem.commons.serialization.GenCodec.ReadFailure
+import com.avsystem.commons.serialization.json.{JsonStringInput, JsonStringOutput}
+import com.avsystem.commons.serialization.{GenCodec, GenKeyCodec}
+import io.udash.rest.openapi._
+import io.udash.rest.raw._
+
+import scala.annotation.implicitNotFound
+
+trait CodecWithStructure[T] {
```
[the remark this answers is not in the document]
ghik: The typeclass is called `RestStructure`. I didn't care much about this trait's name because it's unlikely that it will be used directly.

rest/src/main/scala/io/udash/rest/DefaultRestClient.scala
```diff
@@ -0,0 +1,66 @@
...
+import com.softwaremill.sttp.Uri.QueryFragmentEncoding
+import com.softwaremill.sttp._
+import io.udash.rest.DefaultSttpBackend.backend
+import io.udash.rest.raw._
+
+object DefaultRestClient {
+  @explicitGenerics def apply[RestApi: RawRest.AsRealRpc : RestMetadata](baseUri: String): RestApi =
+    RawRest.fromHandleRequest[RestApi](asHandleRequest(baseUri))
+
+  def asHandleRequest(baseUri: String): RawRest.HandleRequest =
```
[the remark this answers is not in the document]
ghik: There's scaladoc for `HandleRequest` and a [separate documentation section](https://github.com/AVSystem/scala-commons/blob/master/docs/REST.md#handler-function) about it.
I'll add some minimal scaladoc here.

rest/src/main/scala/io/udash/rest/DefaultRestApiCompanion.scala
```diff
@@ -0,0 +1,264 @@
...
+abstract class DefaultRestServerApiCompanion[Real](implicit
+  inst: MacroInstances[DefaultRestImplicits, OpenApiServerInstances[Real]]
+) extends RestServerOpenApiCompanion[DefaultRestImplicits, Real](DefaultRestImplicits)
+
+/**
+  * Base class for companions of REST API traits used for both REST clients and servers.
+  * Injects `GenCodec` and `GenKeyCodec` based serialization and forces derivation of
+  * [[io.udash.rest.openapi.OpenApiMetadata OpenApiMetadata]].
+  */
+abstract class DefaultRestApiCompanion[Real](implicit
```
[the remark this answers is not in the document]
Starzu: Most users will use one of `DefaultRestClientApiCompanion`, `DefaultRestServerApiCompanion` or `DefaultRestApiCompanion`. I think we should left these companions in this file and move everything else to something like `companions.scala`.

rest/src/main/scala/io/udash/rest/DefaultRestClient.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+    * Creates a [[io.udash.rest.raw.RawRest.HandleRequest HandleRequest]] function which sends REST requests to
+    * a specified base URI using default HTTP client implementation (sttp).
+    */
+  def asHandleRequest(baseUri: String): RawRest.HandleRequest =
+    asHandleRequest(uri"$baseUri")
+
+  private def toSttpRequest(baseUri: Uri, request: RestRequest): Request[String, Nothing] = {
+    val uri = baseUri |>
```
ddworak: Can't we use sttp uri interpolation to make this less of an exercise to read?

rest/src/main/scala/io/udash/rest/DefaultRestClient.scala
```diff
@@ -0,0 +1,70 @@
...
+      mimeType => (HeaderNames.ContentType, s"$mimeType;charset=utf-8")
+    }
+    val paramHeaders = request.parameters.headers.iterator.map {
+      case (n, HeaderValue(v)) => (n, v)
+    }.toList
+
+    sttp.copy[Id, String, Nothing](
+      method = Method(request.method.name),
+      uri = uri,
+      headers = contentTypeHeader.toList ++ paramHeaders,
```
ddworak: If order does not matter that much, you can swap these and avoid `.toList`.

rest/src/main/scala/io/udash/rest/DefaultRestClient.scala
```diff
@@ -0,0 +1,70 @@
...
+      ),
+      sttpResp.contentType.fold(HttpBody.empty) { contentType =>
+        val mimeType = contentType.split(";", 2).head
+        HttpBody(sttpResp.body.fold(identity, identity), mimeType)
+      }
+    )
+
+  private def asHandleRequest(baseUri: Uri): RawRest.HandleRequest =
+    RawRest.safeHandle(request => {
+      val sttpReq = toSttpRequest(baseUri, request)
```
ddworak: Inline?

rest/src/main/scala/io/udash/rest/annotations.scala
```diff
@@ -1,49 +1,172 @@
...
+  * By default, parameters of a prefix method are interpreted as URL path fragments. Their values are encoded as
+  * [[PathValue]] and appended to URL path. Alternatively, each parameter may also be explicitly annotated as
+  * [[Header]] or [[Query]].
+  *
+  * NOTE: REST method is interpreted as prefix method by default which means that there is no need to apply [[Prefix]]
+  * annotation explicitly unless you want to specify a custom path.
+  *
+  * @param path see [[RestMethodTag.path]]
+  */
+class Prefix(val path: String = null) extends RestMethodTag
```
ddworak: The docs only mention that it can be empty, not null. Is there a difference? Improve the type or docs.

ghik: Also see: https://github.com/AVSystem/scala-commons/blob/master/docs/Annotations.md

rest/src/main/scala/io/udash/rest/DefaultRestClient.scala
ghik: That will be much slower potentially. `contentTypeHeader.toList` has at most one element so the current version will be O(1).

rest/src/main/scala/io/udash/rest/DefaultRestClient.scala
ghik: Oh why? It looks nice now. Originally I wanted to expose it for reuse.

rest/src/main/scala/io/udash/rest/DefaultRestClient.scala
ghik: Please suggest a solution.

ddworak: I'd argue this is more readable:
```scala
    val pathParams = request.parameters.path.map(_.value)
    val queryParams = request.parameters.query.mapValues(_.value)
    val uri = uri"$baseUri/$pathParams?$queryParams"
```
Passes the tests, no modifications needed.

rest/src/main/scala/io/udash/rest/DefaultRestClient.scala
ddworak: To delay `toSttpRequest` until the callback, should not matter much though.

ghik: Actually, since this request is reusable, my specific intention was to *not* delay it until callback. Why do the same job potentially multiple times? If it throws exception then `RawRest.safeHandle` will take care of this.

ddworak: Convinced after last remark, did not notice that.

rest/src/main/scala/io/udash/rest/DefaultRestClient.scala
ghik: Does it work if `baseUri` already has path suffix and query params?

ddworak: I believe they treat the first interpolated argument specially, but you'd have to add a test case.

ghik:
```scala
val baseUri = "http://localhost:8080/fag/lol/dafuq?foo=bar&kek=lel"
val queryParams = Map("a" -> "b", "c" -> "d")
val pathSegs = List("fuuu", "baar")
println(uri"$baseUri/$pathSegs?$queryParams")
```
gives:
```
http://localhost:8080/fag/lol/dafuq?foo=bar&kek=lel/List(fuuu,+baar)?Map(a+-%3E+b,+c+-%3E+d)
```
😕

ghik: It should be possible to specify URI "prefix", i.e. `"scheme://host:port/basePath"`. I don't care about "base query parameters" and indeed things work well when they are omitted. 

However, in this case we should do some validation so that no abomination like the one above gets generated accidentally. I'm also not sure how the interpolation handles URL encoding.
I must say that with the raw version that I wrote I feel the most confident because it has all these details in it, despite being syntactically inferior.

If you have the patience to rewrite and test this, feel free!

ghik:
What are we doing about the [docs](https://github.com/AVSystem/scala-commons/blob/master/docs/REST.md)?

PR #246, opened 2019-01-14
Ensure element selected in subSeq is Seq

Starzu:

PR #250, opened 2019-02-05
Seq property fixes

mowczare:
Resolves https://github.com/UdashFramework/udash-core/pull/246


mowczare:
You are absolutely right, this PR purpose is only to allow user creating models with all the subclasses of `scala.collection.Seq`.

According to the https://github.com/UdashFramework/udash-core/pull/246 API restriction was not the way to go. It was a decision between model selection elasticity (and accepting possible structure copies) and model restrictions (as in aforementioned PR).

I prefer the elastic approach. Why should we return compilation errors if someone has `scala.collection.mutable.Buffer[T]` in model and want to get `.subSeq`? Same with the immutable equivalent.
That said, I'm open for a discussion.

Starzu:
Could you prepare a benchmark comparing operations on model with `Seq` and `ISeq`?

You've created a lot of compilation related tests, but could you also add some runtime tests?

mowczare:
I've added a benchmark for creating `ModelProperty` with the different type of Seq. As @ghik noticed, obtaining the seq-like value to different models than `Seq` has visible time overhead.

ghik:
Not having too much to do with frontend or Udash development, I don't feel like I can give any meaningful feedback on this.

Starzu:
I think it's ok, but you should add a note in `subSeq` docs and guide.

mowczare:
Added both doc in `subSeq` and paragraph in guide:
https://github.com/UdashFramework/udash-guide/pull/43

@Starzu, please review.

PR #254, opened 2019-02-20
Customizable navbar content

Starzu:


bootstrap4/.js/src/main/scala/io/udash/bootstrap/nav/UdashNavbar.scala
```diff
@@ -87,6 +86,39 @@ object UdashNavbar {
...
+    * More: <a href="http://getbootstrap.com/docs/4.1/components/navbar/">Bootstrap Docs</a>.
+    *
+    * @param expandBreakpoint  Screen size breakpoint to switch between collapsed and expanded menu.
+    * @param darkStyle         If true, enables dark navigation bar theme.
+    * @param backgroundStyle   Selects navigation bar background style.
+    * @param position          Sets bar position.
+    * @param componentId       An id of the root DOM node.
+    * @param navigationFactory Navigation content factory - if you want to clean up the created navigation on
+    *                          the navigation bar cleanup pass it to the provided interceptor.
+    * @param brand             A brand DOM element.
```
FreexD: it's rather a `Modifier`

bootstrap4/.js/src/main/scala/io/udash/bootstrap/nav/UdashNavbar.scala
```diff
@@ -87,6 +86,39 @@ object UdashNavbar {
...
+    * @return A `UdashNavbar` component, call `render` to create a DOM element.
+    */
+  def customContent[ItemType, ElemType <: ReadableProperty[ItemType]](
+    expandBreakpoint: ReadableProperty[BootstrapStyles.ResponsiveBreakpoint] = BootstrapStyles.ResponsiveBreakpoint.Large.toProperty,
+    darkStyle: ReadableProperty[Boolean] = UdashBootstrap.False,
+    backgroundStyle: ReadableProperty[BootstrapStyles.Color] = BootstrapStyles.Color.Light.toProperty,
+    position: ReadableProperty[UdashNavbar.Position] = Position.Auto.toProperty,
+    componentId: ComponentId = ComponentId.newId()
+  )(
+    navigationFactory: Binding.NestedInterceptor => Modifier,
```
FreexD: this is a bit unsafe. Since you later `render` this `Modifier` you should make sure user does not pass anything stupid like `attr` or some `style` as `Modifier` here. I'd go with `Binding.NestedInterceptor => Element`.

Starzu: `Element` == one element - sometimes it's too big restriction and a problem in combination with the bootstrap styles. As you can see `brand` is also a `Modifier` to make it more flexible.

bootstrap4/.js/src/main/scala/io/udash/bootstrap/nav/UdashNavbar.scala
Starzu: The "element" here means something you want to use in a navigation bar, not an instance of `Element` type.

FreexD: Ok.

bootstrap4/.js/src/main/scala/io/udash/bootstrap/nav/UdashNavbar.scala
FreexD: I understand. How about a warning in the documentation comment to warn the user?

PR #260, opened 2019-03-14
Better REST errors

ghik:
* Using `scala-commons` 1.34.10 which supports customization of errors raised from RPC macro engine. See https://github.com/AVSystem/scala-commons/pull/102
* Errors raised for common mistakes in RPC traits are much nicer now. See [test](https://github.com/UdashFramework/udash-core/blob/better-rest-errors/rest/src/test/scala/io/udash/rest/CompilationErrorsTest.scala#L32) for examples.
* Single body parameters are now supported through `@CustomBody` annotation on the **method** rather than `@Body` annotation on the parameter. This is a backwards incompatible change but this feature should be rarely used (so far).
* Introduced `ToAsync` and `FromAsync` typeclasses which make it easier to implement support for other async abstractions (e.g. Monix `Task`).


rest/src/main/scala/io/udash/rest/openapi/OpenApiMetadata.scala
```diff
@@ -166,27 +176,41 @@ case class OpenApiGetOperation[T](
...
 
   def requestBody(resolver: SchemaResolver): Opt[RefOr[RequestBody]] =
-    singleBody.map(_.requestBody(resolver).opt).getOrElse {
-      if (bodyFields.isEmpty) Opt.Empty else Opt {
-        val fields = bodyFields.iterator.map(p => (p.info.name, p.schema(resolver))).toList
-        val requiredFields = bodyFields.collect { case p if !p.info.hasFallbackValue => p.info.name }
-        val schema = Schema(`type` = DataType.Object, properties = Mapping(fields), required = requiredFields)
-        val mimeType = if (formBody) HttpBody.FormType else HttpBody.JsonType
-        RefOr(RestRequestBody.simpleRequestBody(mimeType, RefOr(schema), requiredFields.nonEmpty))
+    if (bodyFields.isEmpty) Opt.Empty else Opt {
```
ddworak: Could be an `.optIf` if you like

rest/src/main/scala/io/udash/rest/openapi/OpenApiMetadata.scala
```diff
@@ -166,27 +176,41 @@ case class OpenApiGetOperation[T](
...
+      val fields = bodyFields.iterator.map(p => (p.info.name, p.schema(resolver))).toList
+      val requiredFields = bodyFields.collect { case p if !p.info.hasFallbackValue => p.info.name }
+      val schema = Schema(`type` = DataType.Object, properties = Mapping(fields), required = requiredFields)
+      val mimeType = bodyTypeTag match {
+        case _: JsonBody => HttpBody.JsonType
+        case _: FormBody => HttpBody.FormType
+        case _ => throw new IllegalArgumentException(s"Unexpected body type $bodyTypeTag")
```
ddworak: `BodyTypeTag` is sealed and only has a couple more implementations, you could explicitly check all cases here to enable exhaustivity verification for future changes.

docs/REST.md
```diff
@@ -364,66 +367,61 @@ Calling `getUsername("ID")` will make a HTTP request on path `users/ID/name`.
...
-unannotated parameters (implicitly interpreted as `@BodyField`s). This way the value of that parameter
-will be serialized straight into HTTP body. This gives you full control over the contents sent in HTTP body
-and their format (i.e. it no longer needs to be `application/json`).
+Non-`GET` methods may be annotated as `@FormBody`. This changes serialization of body parameters
+from JSON object to HTTP form, encoded as `application/x-www-form-urlencoded`. Each body parameter itself
+is then serialized into `QueryValue` rather than `JsonValue`.
+
+#### `@CustomBody`
+
+Methods annotated as `@CustomBody` are required to take **exactly one** body parameter. This body parameter will
```
ddworak: not 100% sure, but `annotated with` is probably more common

docs/REST.md
```diff
@@ -364,66 +367,61 @@ Calling `getUsername("ID")` will make a HTTP request on path `users/ID/name`.
...
 
-### Single body parameters
+#### `@FormBody`
 
-Every non-`GET` API method may also take a single parameter annotated as `@Body` in place of multiple
-unannotated parameters (implicitly interpreted as `@BodyField`s). This way the value of that parameter
-will be serialized straight into HTTP body. This gives you full control over the contents sent in HTTP body
-and their format (i.e. it no longer needs to be `application/json`).
+Non-`GET` methods may be annotated as `@FormBody`. This changes serialization of body parameters
+from JSON object to HTTP form, encoded as `application/x-www-form-urlencoded`. Each body parameter itself
```
ddworak: redundant `itself` (or different word order)

docs/REST.md
```diff
@@ -594,32 +593,30 @@ defined by each "real" type manually or plugged in by REST API trait companion.
...
-failed computation of value of type `T`. It's the most raw, low-level representation of repeatable asynchronous
-computations.
+and translated into `Async[RestResponse]`. This means that the macro engine looks for an implicit instance of 
+`AsRaw[Async[RestResponse], Try[R]]` and `AsReal[Async[RestResponse], Try[R]]` for every HTTP method with result type `R`.
 
-This means that macro engine looks for an implicit instance of `AsRaw/AsReal[Async[RestResponse], Try[R]]`
-for every HTTP method with result type `R`.
+* `Async` is a type alias defined in `RawRest` object.
+  This is just a very raw, low-level representation of asynchronous computations which may be invoked many times. 
+  It serves as a common denominator for all possible asynchronous abstractions like `Future`, Monix `Task`, etc.
```
ddworak: The abstractions are not asynchronous themselves.

docs/REST.md
```diff
@@ -780,66 +779,24 @@ However, `Future` is low level and limited in many ways (e.g. there is no way to
 asynchronous computation starts). It is possible to use other task-like containers, e.g.
 [Monix Task](https://monix.io/docs/2x/eval/task.html) or [Cats IO](https://typelevel.org/cats-effect/).
 
-In order to do that, you must provide some additional "serialization" implicits which will translate your
-wrapped results into `Async[RestResponse]`. Just like when
-[providing serialization for third party type](#providing-serialization-for-third-party-type),
-you should put that implicit into a trait and inject it into REST API trait's companion object.
+In order to do that, you must provide some additional "serialization" implicits which will make the macro engine
+understand how to translate between `Async[T]` and `Task[T]` for arbitrary type `T`. This is controlled by
```
ddworak: Is it translation or conversion?

macros/src/main/scala/io/udash/macros/MiscMacros.scala
```diff
@@ -0,0 +1,27 @@
+package io.udash
+package macros
+
+import com.avsystem.commons.macros.AbstractMacroCommons
+
+import scala.reflect.macros.{TypecheckException, blackbox}
+
+class MiscMacros(val ctx: blackbox.Context) extends AbstractMacroCommons(ctx) {
```
ddworak: Would this work if put into `macros/test/main/...`?

rest/src/main/scala/io/udash/rest/openapi/OpenApiMetadata.scala
ghik: This seems more readable to me. `optIf` would be OK for a smaller expression but I don't really like it for a large block like this one.

docs/REST.md
ghik: These abstractions _do_ capture asynchronicity, don't they?

ddworak: I wouldn't define asynchronous as "captures asynchronicity", but it's just a nitpick ;)

docs/REST.md
ghik: What's the difference? For me, "conversion" implies actual _data_ conversion. In this case it's more like wrapping or adapting, hence "translation".

ddworak: Exactly my thoughts. I just wanted to make sure it's supposed to rather be a wrapper than a converter.

docs/REST.md
ghik: `s/annotated as/annotated with/g`

macros/src/main/scala/io/udash/macros/MiscMacros.scala
ghik: The most annoying thing is that _neither_ ScalaTest nor Shapeless (`illTyped`) do what I need.

docs/REST.md
ghik: Actually I'm curious to hear how you would word it.

ddworak: I'd probably go with `for abstractions over asynchronicity, like` to avoid repetition on `computation`.

PR #261, opened 2019-03-18
REST: bunch of improvements

ghik:
Sorry for the mess - I didn't have the time to split various improvments into separate PRs. Here's a rough list of what was changed:

* `RequestAdjuster` and `ResponseAdjuster` base annotation traits. Methods can be annotated with adjuster annotations which modify outgoing requests and responses in some way (e.g. by adding a header).
* Simplified serialization of path/header/query parameters - replaced `PathValue`, `HeaderValue` and `QueryValue` with a single `PlainValue` assuming that it doesn't really make sense to have different serializations on the same type for path/header/query params. Separate "real" type should be created for such situation.
* Stopped using `com.avsystem.commons.misc.Mapping` and rewritten it in Udash. Separated `Mapping` (case sensitive) from `IMapping` (case insensitive, used for headers). Both these types represent a sequence of key-value pairs where keys may not necessarily be unique + an ability to fetch value by key. In the future we might extend these classes to support multi-valued parameters (e.g. exploded representation of lists in query parameters).
* Replaced `ToAsync` & `FromAsync` with a single `AsyncEffect` bidirectional translation typeclass. `AsRaw` and `AsReal` are separate mainly for better ScalaJS DCE but with `AsyncEffect` that should not be a problem as there should be very few, simple instances of this typeclass at most.
* Some more improvements to compilation error messages
* `RestDataWrapperCompanion` - like `RestDataCompanion` but for transparent wrapper classes. It establishes implicit dependency between wrapper and "wrapee" for `GenCodec`, `GenKeyCodec` and `RestSchema` instances. The biggest benefit from that is that you don't have to manually implement `GenKeyCodec` if it's already available for the wrapped type.


docs/REST.md
```diff
@@ -779,10 +779,11 @@ However, `Future` is low level and limited in many ways (e.g. there is no way to
...
 [Monix Task](https://monix.io/docs/2x/eval/task.html) or [Cats IO](https://typelevel.org/cats-effect/).
 
-In order to do that, you must provide some additional "serialization" implicits which will make the macro engine
+In order to do that, you must provide some additional implicits which will make the macro engine
 understand how to translate between `Async[T]` and `Task[T]` for arbitrary type `T`. This is controlled by
-`ToAsync` and `FromAsync` typeclasses defined in `RawRest` object. This means that you must provide implicit instances
-of `ToAsync[Task]` and `FromAsync[Task]`.
+`AsyncEffect` typeclass defined in `RawRest` object which represents a bidirectional polymorphic conversion between
+some effect type constructor and `Async`. This means that you must provide implicit instance
+of `AsyncEffect[Task]`.
```
ddworak: This change probably affects `REST.svg` as well

rest/src/main/scala/io/udash/rest/raw/AbstractMapping.scala
```diff
@@ -0,0 +1,109 @@
...
+  def nonEmpty: Boolean = entries.nonEmpty
+  def iterator: Iterator[(String, V)] = entries.iterator
+
+  def isDefinedAt(key: String): Boolean =
+    toMap.contains(normKey(key))
+  def apply(key: String): V =
+    toMap(normKey(key))
+
+  override def applyOrElse[A <: String, B >: V](key: A, default: A => B): B =
+    if (caseSensitive) toMap.applyOrElse(key, default)
```
ddworak: Wouldn't calling `normKey` every time do the same? Is it related to performance? Should JIT quite well.

rest/src/main/scala/io/udash/rest/raw/AbstractMapping.scala
```diff
@@ -0,0 +1,109 @@
+package io.udash
+package rest.raw
+
+import com.avsystem.commons._
+import io.udash.rest.raw.AbstractMapping.ConcatIterable
+
+import scala.collection.generic.CanBuildFrom
+import scala.collection.mutable
+
+sealed abstract class AbstractMapping[V] extends PartialFunction[String, V] {
```
ddworak: Probably needs at least a small comment on why we need yet another collection for a kv list

rest/src/main/scala/io/udash/rest/raw/AbstractMapping.scala
```diff
@@ -0,0 +1,109 @@
...
+
+  override def applyOrElse[A <: String, B >: V](key: A, default: A => B): B =
+    if (caseSensitive) toMap.applyOrElse(key, default)
+    else toMap.applyOrElse(normKey(key), (_: String) => default(key))
+
+  def append(key: String, value: V): Self =
+    if (entries.isEmpty) createNew(List((key, value)))
+    else createNew(new ConcatIterable(entries, List((key, value))))
+
+  def prepend(key: String, value: V): Self =
```
ddworak: I'd rather normalize the keys in `append` & `prepend` and not have to keep thinking about accessing the entries via `toMap` everywhere, this is error-prone. Unless you really want to keep the original value for `entries`, but that needs comment as well.

rest/src/main/scala/io/udash/rest/raw/AbstractMapping.scala
```diff
@@ -0,0 +1,109 @@
...
+  * A version of [[Mapping]] which is case-insensitive when looking up values by key.
+  * Used primarily to represent [[io.udash.rest.Header Header]] parameter values.
+  */
+final case class IMapping[V](entries: IIterable[(String, V)]) extends AbstractMapping[V] {
+  type Self = IMapping[V]
+  def caseSensitive: Boolean = false
+
+  protected def createNew(entries: IIterable[(String, V)]): IMapping[V] = IMapping(entries)
+}
+object IMapping extends AbstractMappingCompanion[IMapping]
```
ddworak: How about some tests for these mappings?

rest/src/main/scala/io/udash/rest/raw/AbstractMapping.scala
ghik: There are docs on subclasses, you think it's enough?

rest/src/main/scala/io/udash/rest/raw/AbstractMapping.scala
ghik: Yes, I want to keep original keys as they came. I want only the lookup by key to be case insensitive.

rest/src/main/scala/io/udash/rest/raw/AbstractMapping.scala
ghik: For case-insensitive, I need to rewrap the `default` into yet another lambda, which I wanted to avoid for case-sensitive.

rest/src/main/scala/io/udash/rest/raw/AbstractMapping.scala
ddworak: As long as you don't use `AbstractMapping` in user API it's fine.

PR #268, opened 2019-04-03
RestServlet now has a configurable limit on request payload size

ghik:


rest/.jvm/src/test/scala/io/udash/rest/HttpRestCallTest.scala
```diff
@@ -24,4 +25,10 @@ class HttpRestCallTest extends AbstractRestCallTest with UsesHttpServer {
 
   def clientHandle: HandleRequest =
     SttpRestClient.asHandleRequest(s"$baseUrl/api")
+
+  test("too large binary request") {
+    val future = proxy.binaryEcho(Array.fill[Byte](16 * 1024 * 1024 + 1)(5))
```
ddworak: You test against the default (without reference) here. Use reference to default limit and add a case for custom value - this doesn't check you respect the servlet argument.

PR #278, opened 2019-04-19
Handle anchor without href

### LINKED ISSUE #277: UrlChangeProvider throws error when clicking anchor elements that don't have an href attribute
When clicking an anchor element that does not have an href attribute defined the following error is thrown: 

> RuntimeString.scala:216 Uncaught TypeError: Cannot read property 'indexOf' of null
    at $c_Lio_udash_routing_WindowUrlPathChangeProvider$.io$udash$routing$WindowUrlPathChangeProvider$$$anonfun$initialize$4__Lorg_scalajs_dom_raw_MouseEvent__O (RuntimeString.scala:216)
    at HTMLDocument.x$2 (UrlChangeProvider.scala:93)

I believe the issue comes from this line:

https://github.com/UdashFramework/udash-core/blob/a04920157851bb5c412d3f9c71ebd1cf0c634642/core/.js/src/main/scala/io/udash/routing/UrlChangeProvider.scala#L101

I think adding a null check would fix it but I don't currently have a project set up to rebuild the library from source to test it.

ddworak:
Closes #277

### DIFF
--- a/core/.js/src/main/scala/io/udash/routing/UrlChangeProvider.scala
+++ b/core/.js/src/main/scala/io/udash/routing/UrlChangeProvider.scala
@@ -1,11 +1,11 @@
 package io.udash.routing
 
-import com.avsystem.commons.misc.Opt
+import com.avsystem.commons._
 import io.udash.core.Url
 import io.udash.properties.MutableBufferRegistration
 import io.udash.utils.Registration
 import org.scalajs.dom
-import org.scalajs.dom.raw.HashChangeEvent
+import org.scalajs.dom.raw.{HTMLAnchorElement, HashChangeEvent}
 import org.scalajs.dom.{Element, Location}
 
 import scala.scalajs.js
@@ -31,6 +31,7 @@ trait UrlChangeProvider {
 
 /** Used for routing based on the URL part following # sign. */
 object WindowUrlFragmentChangeProvider extends UrlChangeProvider {
+
   import dom.{document, window}
 
   private val callbacks: js.Array[Url => Unit] = js.Array()
@@ -54,6 +55,7 @@ object WindowUrlFragmentChangeProvider extends UrlChangeProvider {
   * Don't forget to configure your web server to handle frontend routes. You may find "rewrite rules" mechanism useful.
   */
 object WindowUrlPathChangeProvider extends UrlChangeProvider {
+
   import dom.{document, window}
   import org.scalajs.dom.experimental.{URL => JSUrl}
   import org.scalajs.dom.raw.{MouseEvent, Node, PopStateEvent}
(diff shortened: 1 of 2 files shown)


core/.js/src/main/scala/io/udash/routing/UrlChangeProvider.scala
```diff
@@ -91,28 +93,21 @@ object WindowUrlPathChangeProvider extends UrlChangeProvider {
...
-        val ignore = shouldIgnoreClick(event, target, href, samePath, sameHash, sameOrigin)
-
-        if (!ignore) {
-          if (!samePath) {
-            val url = Url(href)
-            changeFragment(url)
+      event.target.opt
+        .collect { case node: Node => node }
+        .flatMap(Iterator.iterate(_)(_.parentNode).takeWhile(_ != null).collectFirstOpt { case a: HTMLAnchorElement => a })
+        .filter(_.href != null)
```
FreexD: what if `href` is an empty string or just `#`?

core/.js/src/main/scala/io/udash/routing/UrlChangeProvider.scala
```diff
@@ -91,28 +93,21 @@ object WindowUrlPathChangeProvider extends UrlChangeProvider {
...
+        .foreach { target =>
+          val href = target.getAttribute("href")
```
FreexD: you can simply `target.href` to get this

core/.js/src/main/scala/io/udash/routing/UrlChangeProvider.scala
```diff
@@ -91,28 +93,21 @@ object WindowUrlPathChangeProvider extends UrlChangeProvider {
...
+          val location = window.location
+          val newUrl = new JSUrl(href, location.toString)
+          val (samePath, sameHash, sameOrigin) =
+            (isSamePath(location, newUrl), isSameHash(location, newUrl), isSameOrigin(location, newUrl))
+          if (!shouldIgnoreClick(event, target, href, samePath, sameHash, sameOrigin)) {
+            if (!samePath) changeFragment(Url(href))
+            event.preventDefault()
```
FreexD: should we also `stopPropagation()` here?

core/.js/src/main/scala/io/udash/routing/UrlChangeProvider.scala
```diff
@@ -132,5 +127,5 @@ object WindowUrlPathChangeProvider extends UrlChangeProvider {
   }
 
   override def currentFragment: Url =
-    Url(Option(window.history.state).map(_.asInstanceOf[js.Dynamic].url.asInstanceOf[String]).getOrElse(window.location.pathname))
+    Url(window.history.state.opt.map(_.asInstanceOf[js.Dynamic].url.toString).getOrElse(window.location.pathname))
```
FreexD: Could `Url` be serialized directly as state data? So that you wouldn't have to recreate it each time you query for current fragment?

ddworak: It shouldn't be created - it's a value class. I'm actually using the URL part of the state itself here, not the state data itself (which seems to be a redundant copy of the same thing). I'll try to remove it from the state data altogether to avoid confusion.

core/.js/src/main/scala/io/udash/routing/UrlChangeProvider.scala
ddworak: Do you see a reason to? It can only prevent user callbacks from execution on this afaik and as a framework we shouldn't prevent elaborate hacks unless we really need to (as is the case with preventing browser default action).

core/.js/src/main/scala/io/udash/routing/UrlChangeProvider.scala
ddworak: Both are correct URL according to HTML5, seem to work fine.

core/.js/src/main/scala/io/udash/routing/UrlChangeProvider.scala
FreexD: After some thought I think it's fine as it is now.

PR #283, opened 2019-05-08
Callback queuing cleanup

### LINKED ISSUE #282: Killed binding fires again
Tested on `0.8.0-RC2`.

```scala
val p = Property(1)
val s = SeqProperty(1,2,3)

div(
  produceWithNested(p) { case (v, nested) =>
    div(
      nested(repeat(s) { v2 =>
        println((v, v2.get))
        div().render
      })
    ).render
  }
).render

CallbackSequencer().sequence {
  p.set(2)
  s.set(Seq(4,5,6))
}
```

The code above prints:
```
(1,1) 
(1,2)
(1,3)
(2,4)
(2,5)
(2,6)
(1,4)
(1,5)
(1,6)
```

The last 3 lines are unexpected. It seems `CallbackSequencer` keeps the listener from `SeqPropertyModifierUtils:108` and the `handlePatch` method does not check if the binding was already killed.

ddworak:
Closes #282

### DIFF
--- a/core/.js/src/test/scala/io/udash/bindings/TagsBindingTest.scala
+++ b/core/.js/src/test/scala/io/udash/bindings/TagsBindingTest.scala
@@ -9,8 +9,10 @@ import io.udash.testing.UdashFrontendTest
 import org.scalajs.dom.Node
 
 import scala.collection.mutable
+import scala.collection.mutable.ListBuffer
 
 class TagsBindingTest extends UdashFrontendTest with Bindings { bindings: Bindings =>
+
   import scalatags.JsDom.all._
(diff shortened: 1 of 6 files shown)


ddworak:
@Starzu it's pretty much the same for 1 listener, faster with copying for 10 listeners (40 vs 80 nanos/op) and faster with set for 100 listeners (2000 vs 5000 nanos/op). I'd say that's a tradeoff, and should still be better than it used to be before, as there was more copying involved in some cases. Set is not that bad for the usual case (< 5 listeners) and scales better when there's a lot of them.

PR #300, opened 2019-05-29
Seq subProp regressions fix

ddworak:
This PR patches the issues introduced by adding `Seq` subtypes support in models.

I've detected 2 major issues so far and fixed them here:
1. `.subProp(_.myVector).get` would lead to a runtime class cast exception
2. `Seq` subtypes with wrong shape wouldn't work (easier to work around when there are just "more" type args, worse when the type was e.g. `scala.xml.Document`, which happens to be an `AbstractSeq[Node]`)

The approach I've taken is to:
- patch `DirectSeqPropertyImpl#get` to map to correct sequence type
- ensure `SeqPropertyCreator` implicit search is used when we're working on `SeqProperties` (as opposed to previous "spray and pray" method (look for a `PropertyCreator` and cast to `SeqPropertyCreator`)
- verify what I can outside of macros with `SeqTpe[A] <:< Seq[A]` and `CanBuildFrom` requirements
- check edge cases in macros `reifySeqPropertyCreator`, `reifySubSeq`, `reifyRoSubSeq`
- cleanup and unify property macro code in regards to validating sequence types

I've tried patching returned subproperties with some delegation magic before this - it also worked, but there were multiple ugly casts/edge cases to consider that I wouldn't remember a week from now.

### DIFF
--- a/macros/src/main/scala/io/udash/macros/PropertyMacros.scala
+++ b/macros/src/main/scala/io/udash/macros/PropertyMacros.scala
@@ -8,29 +8,24 @@ class PropertyMacros(val ctx: blackbox.Context) extends AbstractMacroCommons(ctx
 
   import c.universe._
 
-  val Package = q"_root_.io.udash.properties"
-  val IsModelPropertyTemplateCls = tq"$Package.IsModelPropertyTemplate"
-
-  val PropertyCreatorCls = tq"$Package.PropertyCreator"
-  val SeqPropertyCreatorCls = tq"$Package.SeqPropertyCreator"
-  val SinglePropertyCreatorCls = tq"$Package.SinglePropertyCreator"
-  val PropertyCreatorCompanion = q"$Package.PropertyCreator"
-  val ModelPropertyCreatorCls = tq"$Package.ModelPropertyCreator"
-
-  val ReadablePropertyCls = tq"$Package.single.ReadableProperty"
-  val CastableReadablePropertyCls = tq"$Package.single.CastableReadableProperty"
-  val PropertyCls = tq"$Package.single.Property"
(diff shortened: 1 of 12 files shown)


macros/src/main/scala/io/udash/macros/PropertyMacros.scala
```diff
@@ -314,32 +321,36 @@ class PropertyMacros(val ctx: blackbox.Context) extends AbstractMacroCommons(ctx
...
-        c.abort(c.enclosingPosition,
-          s"""
-             |The path must consist of ModelProperties and only leaf can be a Property, ModelProperty or SeqProperty.
-             | * ${next.tpe.widen} ${if (hasModelPropertyCreator(next.tpe.widen)) "is" else "is NOT"} a ModelProperty
-             | * $t ${if (isValidSubproperty(next.tpe, t)) "is" else "is NOT"} a valid subproperty (abstract val/def for trait based model or constructor element for (case) class based model)
-             |
+    def checkIfIsValidPath(tree: Tree, validation: Type => Unit = _ => ()): Boolean = {
+      tree match {
+        case Select(next@Ident(_), t) if isValidSubproperty(next.tpe, t) =>
+          validation(next.tpe.member(t).typeSignature.finalResultType)
```
Starzu: `validation` -> `abortIfInvalid`? It looks like ignored check and nothing suggests it has a side-effect.

ddworak: Validation is usually side-effectful (see property validation :rofl:). I'd say the return type `Unit` says a lot as well. This does not always have to abort from the perspective of this method - it may just be some logging etc. I'll switch to `onLeafType`

PR #320, opened 2019-06-19
Use coursier for fetching plugins as well

ddworak:


project/project/plugins.sbt
```diff
@@ -0,0 +1 @@
+addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.1.0-M13-4")
```
bgrochal: `M14-4` is recommended in [docs](https://get-coursier.io/docs/sbt-coursier) now

PR #331, opened 2019-06-27
FileUploader allow single file upload without []

smiklos:
I implemented a solution using udash where I upload to S3 using pre-signed urls. 
This method however enforces only one file per post request and the file has to be named `file`. 

I can easily get one file out of my model but with this change I can also support my use case while keeping the original Api intact. Looking at the fileupload servlet, this change shouldn't cause any issues there either.

### DIFF
--- a/core/.js/src/main/scala/io/udash/utils/FileUploader.scala
+++ b/core/.js/src/main/scala/io/udash/utils/FileUploader.scala
@@ -8,6 +8,7 @@ import org.scalajs.dom.raw.{FormData, XMLHttpRequest}
 import scala.scalajs.js
 
 class FileUploader(url: Url) {
+
   import FileUploader._
 
   /** Uploads files selected in provided `input`. */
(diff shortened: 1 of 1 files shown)


smiklos:
Your proposal works even better, I wasn't sure how much you like to keep things backwards compatible. 

On the frontend side, it'd be still great to be able to model a single file upload as I think that's more common than multiple files with the same name. Does it worth making a fileInput with one file property only?

ddworak:
We're happy to change the API before we hit `1.x`, there's still some things we want to polish.

I think there could be more factories in `io.udash.bindings.inputs.FileInput` (e.g. `single` and `multi`), which would forward to the original `apply`. Please note you can easily transform between `Property` and `SeqProperty` for the single case like:
```scala
def single(fileProperty: Property[File])(inputName: String, inputModifiers: Modifier*): InputBinding[JSInput] = {
  apply(fileProperty.transformToSeq[File](Seq(_), _.head), false.toProperty)(inputName, inputModifiers)
}
```

smiklos:
Should I make those changes? do you want to keep a the single file uploader method and just delegate it to the original with file wrapped in Seq()?

ddworak:
If you want to then yes, please :)

As far as the implementation is concerned, the more code reused the better. Wrapping in `Seq` should be an unnoticable overhead in this case.

ddworak:
Closing in favour of https://github.com/UdashFramework/udash-core/pull/336

PR #337, opened 2019-06-28
Add committer list generation script

ddworak:


scripts/authors.scala
```diff
@@ -0,0 +1,70 @@
...
+  updatedEntry.fold(
+    map
+  )(entry => map + (entry.name.toLowerCase -> entry))
+}
+
+val sorted = map.values.toSeq.sortBy(s => (s.commits, s.inserts + s.deletes)).reverse
+
+println("commits  added  removed")
+sorted.foreach { entry =>
+  println("%7d%7d%9d %s".format(entry.commits, entry.inserts, entry.deletes, entry.name))
```
mowczare: You may consider printing email here as well.

scripts/authors.scala
```diff
@@ -0,0 +1,70 @@
...
+ *    scala authors.scala tag1 tag2
+ *
+ * requires scala 2.13.x+ and command line git on path
+ */
+
+import scala.sys.process._
+
+require(args.length == 2, "usage: authors prevTag currTag")
+
+val gitCmd = "git log --no-merges --shortstat -z --minimal -w -C " + args(0) + ".." + args(1)
```
mowczare: Funny thing is, when you call this git command with the `currTag` first and the `prevTag` second - it still works and returns invalid data (commits before the `prevTag`).

Simple `git tag` may be used to validate if the first tag was before the second.

ddworak: `git log` doesn't seem to care so neither will I, it's just a tool for us

PR #350, opened 2019-07-23
Fix produce updates on SeqFrag / DocumentFragments

ddworak:


core/.js/src/main/scala/io/udash/bindings/modifiers/ValueModifier.scala
```diff
@@ -21,11 +24,19 @@ trait ValueModifier[T] extends Binding with DOMManipulator {
...
-      elements = {
-        if (checkNull && propertyValue == null) emptyStringNode()
-        else builder(propertyValue, nestedInterceptor)
+      val newEls: Seq[Node] =
+        builder(propertyValue, nestedInterceptor)
+          .optIf(!checkNull || propertyValue != null)
+          .filter(_.nonEmpty)
+          .getOrElse(emptyStringNode())
+
+      elements = if (newEls.size == 1 && newEls.head.isInstanceOf[DocumentFragment]) {
```
bgrochal: I'm afraid this condition is not sufficient. Please consider your slightly modified test case:
```scala
val p = SeqProperty.blank[String]
val template = div(produce(p)(s => Seq(
  "test".render,
  s.render
))).render
template.outerHTML shouldBe "<div>test</div>"

p.set(Seq("A", "B", "C"))
template.outerHTML shouldBe "<div>testABC</div>"
```
which results in:
```
[info] - should handle SeqFrag update *** FAILED ***
[info]   scala.scalajs.js.JavaScriptException: NotFoundError: Node was not found
```

core/.js/src/main/scala/io/udash/bindings/modifiers/ValueModifier.scala
```diff
@@ -21,11 +24,19 @@ trait ValueModifier[T] extends Binding with DOMManipulator {
...
+        //DocumentFragments nodes are moved after appending, so we cannot use newEls as oldEls later on
+        // (as childNodes mutate to empty). Therefore we need to use a shallow copy directly.
+        //These are often generated by `scalatags.JsDom.Cap.SeqFrag`
+        js.Dynamic.global.Array.prototype.slice.call(newEls.head.childNodes).asInstanceOf[js.Array[Node]]: Seq[Node]
```
bgrochal: What if `newEls.head.childNodes` contains another `DocumentFragment`?

core/.js/src/test/scala/io/udash/bindings/TagsBindingTest.scala
```diff
@@ -776,6 +776,28 @@ class TagsBindingTest extends UdashFrontendTest with Bindings { bindings: Bindin
       template2.textContent should be("")
       oldCounter should be(4)
     }
+
```
bgrochal: I believe we should add test cases for the two aforementioned issues.

core/.js/src/main/scala/io/udash/bindings/modifiers/ValueModifier.scala
```diff
@@ -14,18 +17,28 @@ trait ValueModifier[T] extends Binding with DOMManipulator {
   protected def checkNull: Boolean
   protected def listen(callback: T => Unit): Registration
 
+  private def childrenCopy(node: Node): Iterator[Node] = {
```
bgrochal: Could be local to `applyTo`, or even `rebuild` if you wish,

PR #356, opened 2019-08-02
REST: support for comma-separated, repeated and flag `@Query` params

ghik:
Primary motivation: to allow `List`s and other collections to be used as types of `@Query` parameters

Details:
* `@Query` parameters now serialize to new raw type, `RawQueryValue` instead of `PlainValue`
* `RawQueryValue` allows emitting query strings previously impossible to express:
  * comma separated lists of URL-encoded values, e.g. `ids=1,2,3` (separating commas are *not* URL-encoded but commas appearing in values *are*)
  * repeated parameters, e.g. `id=1&id=2&id=3` (assuming they're not interspersed with other params)
  * parameters without the `=`, e.g. `?enabled` instead of `?enabled=true`
  * binary URL-encoded values which are not valid UTF-8 strings after decoding
* by default, `RawQueryValue` serialization reuses `PlainValue` serialization
* `RawQueryValue` serialization is also provided for any `Iterable` (or subclass) of element `T` given that `PlainValue` serialization is available for `T` - this serialization uses comma-separated representation
* `@explode` annotation may be used to change comma-separated representation into repeated parameters
* `@flag` annotation may be used on `Boolean` query params to use `?param` instead of `?param=true`

The changes above required two new features in RPC framework:
* `EncodingInterceptor` and `DecodingInterceptor` for defining annotations which change serialization of annotated parameter/method
* `@rawWhenAbsent` annotation for providing raw (serialized) default parameter values

### DIFF
--- a/rest/src/main/scala/io/udash/rest/SttpRestClient.scala
+++ b/rest/src/main/scala/io/udash/rest/SttpRestClient.scala
@@ -3,8 +3,6 @@ package rest
 
 import com.avsystem.commons._
 import com.avsystem.commons.annotation.explicitGenerics
-import com.softwaremill.sttp.Uri.QueryFragment.Plain
-import com.softwaremill.sttp.Uri.QueryFragmentEncoding
 import com.softwaremill.sttp._
 import io.udash.rest.raw._
 
@@ -30,14 +28,9 @@ object SttpRestClient {
     asHandleRequest(uri"$baseUri")
 
   private def toSttpRequest(baseUri: Uri, request: RestRequest): Request[Array[Byte], Nothing] = {
-    val uri = baseUri |>
-      (u => u.copy(path = u.path ++
-        request.parameters.path.map(_.value))) |>
-      (u => u.copy(queryFragments = u.queryFragments ++
-        request.parameters.query.entries.iterator.map {
-          case (k, rqv) => Plain(rqv.encodeParam(k), QueryFragmentEncoding.Relaxed)
-        }.toList
-      ))
+    val tmpUri = uri"${request.parameters.toUri(s"tmp://tmp")}"
+    val uri = baseUri |> (u =>
+      u.copy(path = u.path ++ tmpUri.path, queryFragments = u.queryFragments ++ tmpUri.queryFragments))
 
     val contentHeaders = request.body match {
       case HttpBody.Empty => Nil
--- a/rest/src/main/scala/io/udash/rest/SttpRestClient.scala
+++ b/rest/src/main/scala/io/udash/rest/SttpRestClient.scala
@@ -3,7 +3,7 @@ package rest
 
 import com.avsystem.commons._
 import com.avsystem.commons.annotation.explicitGenerics
-import com.softwaremill.sttp.Uri.QueryFragment.KeyValue
+import com.softwaremill.sttp.Uri.QueryFragment.Plain
 import com.softwaremill.sttp.Uri.QueryFragmentEncoding
 import com.softwaremill.sttp._
 import io.udash.rest.raw._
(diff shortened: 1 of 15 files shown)


rest/src/main/scala/io/udash/rest/SttpRestClient.scala
```diff
@@ -30,14 +28,9 @@ object SttpRestClient {
...
-      (u => u.copy(path = u.path ++
-        request.parameters.path.map(_.value))) |>
-      (u => u.copy(queryFragments = u.queryFragments ++
-        request.parameters.query.entries.iterator.map {
-          case (k, PlainValue(v)) => KeyValue(k, v, QueryFragmentEncoding.All, QueryFragmentEncoding.All)
-        }.toList
-      ))
+    val tmpUri = uri"${request.parameters.toUri(s"tmp://tmp")}"
+    val uri = baseUri |> (u =>
+      u.copy(path = u.path ++ tmpUri.path, queryFragments = u.queryFragments ++ tmpUri.queryFragments))
```
ddworak: Not tested well enough.
```scala
      u.copy(path = u.path ++ tmpUri.path, queryFragments = tmpUri.queryFragments))
```
passes the tests.

rest/src/main/scala/io/udash/rest/raw/RawQueryValue.scala
```diff
@@ -0,0 +1,183 @@
...
+      case Single(v) =>
+        _next = null
+        v
+      case Repeated(v, tail) =>
+        _next = tail
+        v
+    }
+  }
+
+  def toList: List[Opt[String]] = iterator.toList
```
ddworak: Unused, can be easily done on callsite.

rest/src/main/scala/io/udash/rest/raw/RawQueryValue.scala
```diff
@@ -0,0 +1,183 @@
...
+  def plain(value: String): RawQueryValue =
+    Single(Opt(urlEncode(value)))
+
+  /**
+    * @param encoded If non-empty, contains an already URL-encoded string (it may contain unencoded commas to indicate
+    *                a list of values). `Opt.Empty` indicates a "flag" query parameter (without `=` and value).
+    */
+  // not using plain List in order to statically ensure non-emptiness and avoid too much boxing of Opts
+  case class Single(encoded: Opt[String]) extends RawQueryValue
+  case class Repeated(encoded: Opt[String], tail: RawQueryValue) extends RawQueryValue
```
ddworak: Both `final` :rofl:

rest/src/main/scala/io/udash/rest/raw/RawQueryValue.scala
```diff
@@ -0,0 +1,183 @@
...
+  def unexploded: RawQueryValue =
+    Single(Opt(iterator.flatten.mkString(",")))
+
+  def firstPlainValue: PlainValue =
+    PlainValue(first.fold("")(urlDecode))
+}
+object RawQueryValue {
+  def fromRaw(values: TraversableOnce[Opt[String]]): RawQueryValue =
+    if (values.isEmpty) throw new IllegalArgumentException("no values")
+    else values.foldRight(null: RawQueryValue) {
```
ddworak: `foldRight` here requires materializing an intermediate reversed collection. Can you enforce order at other level or ensure access to reverse iterator?

rest/src/main/scala/io/udash/rest/raw/RawQueryValue.scala
```diff
@@ -0,0 +1,183 @@
...
+    URLEncoder.encode(str, spaceAsPlus = true)
+
+  def urlDecode(str: String): String =
+    URLEncoder.decode(str, plusAsSpace = true)
+
+  def encodeQuery(query: Mapping[RawQueryValue]): String =
+    encodeQuery(query, new StringBuilder).result()
+
+  def encodeQuery(query: Mapping[RawQueryValue], builder: StringBuilder): StringBuilder = {
+    val it = query.entries.iterator
```
ddworak: `query.iterator`

rest/src/main/scala/io/udash/rest/SttpRestClient.scala
ghik: I actually wonder how much should I care about it. Maybe I should just reject base URI with query parameters.

ddworak: Well, someone might need this obscure feature to pass a parameter along all requests. It wouldn't be easy to do with the framework otherwise I guess.

ddworak:
Closing as stale

PR #369, opened 2019-08-12
Tempus Dominus date picker fixes

bgrochal:
Done:
 - date picker options and listeners are initialized on DOM append to enable caching component's state,
 - single `datetimepicker` call with both options and default date (avoiding stack overflow errors),
 - minor refactoring.

### DIFF
--- a/bootstrap4/.js/src/main/scala/io/udash/bootstrap/datepicker/UdashDatePicker.scala
+++ b/bootstrap4/.js/src/main/scala/io/udash/bootstrap/datepicker/UdashDatePicker.scala
@@ -13,7 +13,8 @@ import io.udash.css.{CssStyle, CssStyleName}
 import io.udash.i18n.{LangProperty, TranslationKey0, TranslationProvider}
 import io.udash.logging.CrossLogging
 import io.udash.wrappers.jquery._
-import org.scalajs.dom.Element
+import org.scalajs.dom.raw.{MutationObserver, MutationObserverInit}
+import org.scalajs.dom.{Element, document}
 
 import scala.scalajs.js
 import scala.scalajs.js.|
@@ -37,6 +38,24 @@ final class UdashDatePicker private[datepicker](
   ).render
   private val jQInput = jQ(inp).asInstanceOf[UdashDatePickerJQuery]
 
+  locally {
+    // When a date picker gets appended to a DOM, its options and listeners should be initialized once again. Thus, all
+    // nodes with mutated children are examined whether there is a date picker defined within a corresponding DOM
+    // subtree. It is also substantially possible for any date picker to be reported by more than one `MutationRecord`
+    // since the `subtree` switch of `MutationObserverInit` is on, thus only the first update is considered (note
+    // `findOpt`) to avoid double initialization.
+    new MutationObserver((records, _) =>
+      records
+        .iterator
+        .flatMap(record => for {i <- 0 until record.addedNodes.length} yield record.addedNodes(i))
+        .findOpt {
+          case element: Element => element.querySelector(s"#$componentId") != null
+          case _ => false
+        }
+        .foreach(_ => setupDatePicker())
+    ).observe(document.body, MutationObserverInit(childList = true, subtree = true))
+  }
+
   /** Shows date picker widget. */
   def show(): Unit =
     jQInput.datetimepicker("show")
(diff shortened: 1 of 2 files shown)


bootstrap4/.js/src/main/scala/io/udash/bootstrap/datepicker/UdashDatePicker.scala
```diff
@@ -37,6 +38,24 @@ final class UdashDatePicker private[datepicker](
...
+    new MutationObserver((records, _) =>
+      records
+        .iterator
+        .flatMap(record => for {i <- 0 until record.addedNodes.length} yield record.addedNodes(i))
+        .findOpt {
+          case element: Element => element.querySelector(s"#$componentId") != null
+          case _ => false
+        }
+        .foreach(_ => setupDatePicker())
+    ).observe(document.body, MutationObserverInit(childList = true, subtree = true))
```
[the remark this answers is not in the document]
bgrochal: Unfortunately we are not able to narrow down `target` because we need to setup a mutation observer **before** appending nodes to be watched to DOM. Since we operate on a generic component here, existence of any parent node cannot be assumed except for the "static" ones (such as `html` and `body`).

I think single mutation observer with `querySelector` is fair enough - such a solution is recommended by many experienced frontend devs.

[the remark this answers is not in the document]
ddworak: How about a "hacky" (yet less involving on our code side) solution for node adding/removal detection like: https://davidwalsh.name/detect-node-insertion ?

bootstrap4/.js/src/main/scala/io/udash/bootstrap/datepicker/UdashDatePicker.scala
```diff
@@ -37,6 +37,38 @@ final class UdashDatePicker private[datepicker](
...
+      jQInput.datetimepicker(
+        optionsToJsDict(options.get)
+          .setup(optionsDict => date.get.foreach(date => optionsDict.update("date", dateToMoment(date))))
+      )
+
+      nestedInterceptor(new JQueryOnBinding(jQInput, "change.datetimepicker", (_: Element, event: JQueryEvent) => {
+        val dateOption = event.asInstanceOf[DatePickerChangeJQEvent].option
+          .flatMap(ev => sanitizeDate(ev.date))
+          .map(momentToDate)
+        val oldDateOption = date.get
```
ddworak: Is this safe?

bootstrap4/.js/src/main/scala/io/udash/bootstrap/datepicker/UdashDatePicker.scala
```diff
@@ -57,46 +89,17 @@ final class UdashDatePicker private[datepicker](
...
+  override val render: Element = {
+    propertyListeners += options.listen(opts =>
+      optionsToJsDict(opts).foreach { case (optionKey, optionValue) => jQInput.datetimepicker(optionKey, optionValue) }
+    )
+    propertyListeners += date.listen(optionalDate => optionalDate.foreach(date => jQInput.datetimepicker("date", dateToMoment(date))))
     inp
   }
 
   override def kill(): Unit = {
     super.kill()
```
ddworak: Usual order is to first kill your stuff, then parent/super. This is due to the fact that `super` is initialized before `this` and some of your stuff might depend on it, hence the stack cleanup order.

bootstrap4/.js/src/main/scala/io/udash/bootstrap/datepicker/UdashDatePicker.scala
```diff
@@ -531,4 +534,37 @@ object UdashDatePicker {
     def valueOf(): Double = js.native
   }
 
+  import org.scalajs.dom.raw.{MutationObserver, MutationObserverInit, MutationRecord}
+  import scala.collection.mutable.{Set => MSet}
+
+  private val datePickerSetupCallbacks = MSet.empty[(ComponentId, () => Unit)]
```
ddworak: Would be cleaner and easier to handle as a map.

bootstrap4/.js/src/main/scala/io/udash/bootstrap/datepicker/UdashDatePicker.scala
```diff
@@ -531,4 +534,37 @@ object UdashDatePicker {
...
+        if (addedNodes.exists {
+          case element: Element => element.querySelector(s"#$pickerId") != null
+          case _ => false
+        }) callback()
+      }
+    })
+
+  def registerSetupCallback(id: ComponentId, callback: () => Unit): Unit = {
+    datePickerSetupCallbacks += ((id, callback))
+    if (datePickerSetupCallbacks.isEmpty)
```
ddworak: I don't understand, is this ever called? You've just added a callback line above.

bootstrap4/.js/src/main/scala/io/udash/bootstrap/datepicker/UdashDatePicker.scala
bgrochal: I believe it is - this `get` is invoked on a `RedableProperty`, not an `Option`.

bootstrap4/.js/src/main/scala/io/udash/bootstrap/datepicker/UdashDatePicker.scala
bgrochal: That's the root cause for failing tests as well, nice catch 👍

bootstrap4/.js/src/main/scala/io/udash/bootstrap/datepicker/UdashDatePicker.scala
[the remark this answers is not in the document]
bgrochal: Mutation Observer approach left because of [better benchmark results](https://output.jsbin.com/hibeji).

PR #371, opened 2019-08-13
FontAwesome 5.10.1

bgrochal:
Things done:
 - FontAwesome 5.10.1 integration,
 - sbt `copyAssets` task configured to copy `webfonts` directory from a FontAwesome WebJar to appropriate `assets` subdirectory (statically provided font files are no longer required, it is sufficient to bump `FontAwesome` dependency version now),
 - legacy `guide/project` directory deleted.


build.sbt
```diff
@@ -443,6 +443,19 @@ lazy val `guide-guide` =
     Dependencies.guideJsDeps,
     Some((`guide-backend`, "io.udash.web.styles.GuideCssRenderer")),
     Def.task(Some((`guide-commons` / sourceDirectory).value / "main" / "assets"))
+  ).settings(
+    Compile / copyAssets := {
```
ddworak: I'd rather you add this logic to the existing `copyAssets` logic in `frontendExecutable` - FontAwesome is available on both frontends anyway via `guide-commons`.

build.sbt
```diff
@@ -443,6 +443,19 @@ lazy val `guide-guide` =
...
+      // (Compile / copyAssets) logic already defined within `frontendExecutable` should be executed first
+      (Compile / copyAssets).value
+      // font-awesome WebJar is required on the classpath to execute the next step
+      (Assets / WebKeys.webJars).value
+      // copying font-awesome webfonts to a location required by font-awesome styles 
+      // (appropriate CSS file is inlined by less within UdashStatics/WebContent/guide/assets/styles/assets.min.css)
+      IO.copyDirectory(
+        (Assets / WebKeys.webJarsDirectory).value / "lib/font-awesome/webfonts",
```
ddworak: I believe `lib` here comes from `webModulesLib` (could change)

build.sbt
```diff
@@ -443,6 +443,19 @@ lazy val `guide-guide` =
...
+        target.value / "UdashStatics/WebContent/guide/assets/webfonts"
```
ddworak: You could ensure this is the same `assets` dir as in other definitions (will be much easier once it's integrated into regular `copyAssets`)

bootstrap4/.js/src/main/scala/io/udash/bootstrap/utils/UdashIcons.scala
```diff
@@ -1404,6 +1598,7 @@ object UdashIcons {
       def wrench: CssStyle = fas("wrench")
       def xRay: CssStyle = fas("x-ray")
       def yenSign: CssStyle = fas("yen-sign")
+      def yinYang: CssStyle = fas("yin-yang")
     }
```
ddworak: Do you think we could add a test for rendering these based on e.g. second example `https://developer.mozilla.org/en-US/docs/Web/API/FontFaceSet/check#Examples`?

bootstrap4/.js/src/test/scala/io/udash/bootstrap/utils/UdashIconsTest.scala
```diff
@@ -1,15 +1,21 @@
...
+import io.udash.testing.AsyncUdashCoreFrontendTest
+import org.scalajs.dom._
+import org.scalatest.Succeeded
+import scalatags.JsDom.all._
+
+import scala.concurrent.Future
+
+class UdashIconsTest extends AsyncUdashCoreFrontendTest with CssView {
   "Bootstrap icons" should {
+    document.body.appendChild(UdashBootstrap.loadFontAwesome())
```
ddworak: I'd rather this was in `beforeAll`, no guarantees about when the constructor is called here at all

bootstrap4/.js/src/main/scala/io/udash/bootstrap/utils/UdashIcons.scala
bgrochal: According to the docs, the two-arg variant of this method

> does not check for individual glyph coverage

([explanation](http://lists.w3.org/Archives/Public/www-style/2015Aug/0330.html)). However I've finally managed to implement a test for FontAwesome API/statics compatibility, please check it out under `UdashIconsTest`.

PR #381, opened 2019-08-28
Support dynamic dropdown items

### LINKED ISSUE #379: Regression in Dropdown API
I'm migrating from `0.8.0-RC5` to `0.8.0`. https://github.com/UdashFramework/udash-core/commit/79783e21a362445506a641e35c48f739389adb57 changed the `itemFactory` type from `(ElemType, Binding.NestedInterceptor) => Element` to `ItemType => Element`, so it's no longer possible to dynamically change dropdown's element content.

ddworak:
The initial idea behind https://github.com/UdashFramework/udash-core/commit/79783e21a362445506a641e35c48f739389adb57 was to avoid additional `<span>` element polluting the DOM in the typical, non-dynamic case. My proposal for the middle-ground between the original version and the current one is to introduce limited support which would push the responsibility for creating a wrapping `Element` to the API user.

Closes #379

### DIFF
--- a/bootstrap4/.js/src/main/scala/io/udash/bootstrap/dropdown/UdashDropdown.scala
+++ b/bootstrap4/.js/src/main/scala/io/udash/bootstrap/dropdown/UdashDropdown.scala
@@ -9,7 +9,7 @@ import io.udash.bootstrap.button.UdashButton
 import io.udash.bootstrap.utils.{BootstrapStyles, UdashBootstrapComponent}
 import io.udash.properties.seq
 import io.udash.properties.single.ReadableProperty
-import io.udash.wrappers.jquery.JQuery
+import io.udash.wrappers.jquery._
 import org.scalajs.dom.{Element, Event}
 import scalatags.JsDom.all._
 
@@ -22,7 +22,7 @@ final class UdashDropdown[ItemType, ElemType <: ReadableProperty[ItemType]] priv
   buttonToggle: ReadableProperty[Boolean],
   override val componentId: ComponentId
 )(
-  itemFactory: (ItemType, Binding.NestedInterceptor) => Element,
+  itemBindingFactory: UdashDropdown[ItemType, ElemType] => Binding,
   buttonContent: Binding.NestedInterceptor => Modifier,
   buttonFactory: (NestedInterceptor => Modifier) => UdashButton
 ) extends UdashBootstrapComponent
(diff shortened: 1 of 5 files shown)


bootstrap4/.js/src/main/scala/io/udash/bootstrap/dropdown/UdashDropdown.scala
```diff
@@ -22,7 +22,7 @@ final class UdashDropdown[ItemType, ElemType <: ReadableProperty[ItemType]] priv
   buttonToggle: ReadableProperty[Boolean],
   override val componentId: ComponentId
 )(
-  itemFactory: ItemType => Element,
+  itemFactory: (ItemType, Binding.NestedInterceptor) => Element,
```
Starzu: This version does not allow dynamic changes of dropdown's elements without rendering the whole menu again. What about extracting the line 81 to constructor as `Binding`? This way you could provide both factory types (`(ElemType, Binding.NestedInterceptor) => Element` and `ItemType => Element`).

bootstrap4/.js/src/main/scala/io/udash/bootstrap/dropdown/UdashDropdown.scala
```diff
@@ -78,7 +78,7 @@ final class UdashDropdown[ItemType, ElemType <: ReadableProperty[ItemType]] priv
         nestedInterceptor(BootstrapStyles.Dropdown.menuRight.styleIf(rightAlignMenu)),
         aria.labelledby := buttonId, id := menuId
       )(nestedInterceptor(
-        produce(items.transform((item: ItemType) => withSelectionListener(itemFactory(item), item)))(el => el)
+        produce(items.transform((item: ItemType) => withSelectionListener(itemFactory(item, nestedInterceptor), item)))(el => el)
```
Starzu: You should use `produceWithNested` to properly clean up bindings in case of `items` changes,

bootstrap4/.js/src/main/scala/io/udash/bootstrap/dropdown/UdashDropdown.scala
ddworak: I was considering it, but that would leave a lot more responsibility to the user, especially in terms of applying the selection listener. With current approach you're free to change the content of a `DefaultDropdownItem.Dynamic` and it shouldn't trigger a full menu render (as the function reference wouldn't change). The only remaining issue would be rerendering on structure change: this can only be solved via `repeat`, which would require nesting an element again. Any suggestions?

Starzu: You could take `Binding` as an argument in this private constructor and expose two `apply` versions in the companion object with these two factory types. Inside you can hide implementation with `produce` and `repeat`.

PR #388, opened 2019-09-02
Travis publish

### LINKED ISSUE #312: Automatically publish new versions
Similarly to how `scala-commons` does it with Travis

ddworak:
Example run (without `sonatypeRelease`): https://travis-ci.org/UdashFramework/udash-core/builds/578854261

Closes #312


.travis.yml
```diff
@@ -16,34 +13,57 @@ before_script:
...
+    - secure: JfVPHMqEITjrDqHbVkQZCR4VLtLUFchunlea7+JoIWpQ2o0RuSUA7jUk+bBG+Avj6qwYSu+1/wmTQY/xRQBTEDTiuf8CKROIX0sPeyTvTcGm31V0z+Ajxu7x4xmLH2hr4/9y35oAwqy0knBK84J3VZ6xGRKO3JQAnk/iotpb2uGj6wjso2rQnG56NNVuM/WIZ5BwRtJrCP7rbyOJTMMT639+ECrmppsEZnODqgARhe5PUT+DRn5MXvH0sayt38SukUkahD5qcnMthP3jbNkaMrZ0RlIAzbOQUvtzt1rwHjeo71V2Dd6q3kLworbkmbmmCbxd7DtWAmTcMir7+91hXLP6r+3q2YFh5juqXXFEDjox0Nqt1nl6sfHAJMEOIuji7XoJLSAK5nOY5OsFZ5WK/yF7JlllaBUFGIJOeQbf4k70Nr6nmer7iHrvaD3H2jhtkyp5cET/0hMNpo7jSnvoAk5l0eZhleWbTgwzbHpjjw/yDPKS+S+q+2DmtTrHcDa1Z79YHps/9hlryGM3kT3mCr2cwFP+aoPnXKGdPuiY95LcVPHeI8TKIopU3lOMXEPX1fTY/xNDauwl9zZfKm2dwJU9Qa3uF5NSgr0s0nzUuj4+VJlTr6tLREAEznPESTwmm9qlHrJPX4aXEDCxRdfgSRFsGoRTqkvoiwozdv8+Z1c=
 
 matrix:
   exclude:
     - scala: 2.11.12
       env: TEST_PACKAGE="guide-selenium/test"
 
+jobs:
+  include:
+    - if: tag =~ ^v
```
ghik: You really want to have separate jobs for different scala versions? This way you lose atomicity of this operation which may cause two problems:
* `sonatypeRelease` from one job may accidentally promote repository uploaded by another one
* you may end up with a partial release when one scala version succeeds but another one fails

ddworak: I don't, but a single job fails due to Travis time limits. I might have to do the release part manually then :(

ghik: That sort of misses the entire point...

ddworak: I attempted:
- caching the targets - Scala usually doesn't care too much anyway, requires exactly same env variables for Travis to pick it up as well
- parallel upload to a staging repository (`sbt-sonatype` 3.1) - works nicely, but messy to encode and error-prone
- bundle uploading (`sbt-sonatype` 3.4) - this seems to be the best solution, it's fast enough to publish Scala 2.11 and 2.12 in 15 minutes without any crazy caching: https://travis-ci.org/UdashFramework/udash-core/jobs/581824091

PR #389, opened 2019-09-02
Scalatags update

ddworak:
Changes required due to: https://github.com/lihaoyi/scalatags/pull/182

### DIFF
--- a/css/src/main/scala/io/udash/css/CssText.scala
+++ b/css/src/main/scala/io/udash/css/CssText.scala
@@ -12,6 +12,6 @@ trait CssText {
 object CssText extends CssText {
   private final class TextStyleModifier(styles: CssStyle*) extends Modifier {
     override def applyTo(t: Builder): Unit =
-      styles.foreach { s => t.appendAttr(cls.name, Builder.GenericAttrValueSource(s.classNames.mkString(" ", " ", ""))) }
+      styles.foreach { s => t.appendAttr(cls.name, Builder.GenericAttrValueSource(s.classNames.mkString)) }
   }
 }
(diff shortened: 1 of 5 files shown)


css/src/main/scala/io/udash/css/CssText.scala
```diff
@@ -12,6 +12,6 @@ trait CssText {
 object CssText extends CssText {
   private final class TextStyleModifier(styles: CssStyle*) extends Modifier {
     override def applyTo(t: Builder): Unit =
-      styles.foreach { s => t.appendAttr(cls.name, Builder.GenericAttrValueSource(s.classNames.mkString(" ", " ", ""))) }
+      styles.foreach { s => t.appendAttr(cls.name, Builder.GenericAttrValueSource(s.classNames.mkString)) }
```
bgrochal: `classNames` should be separated by a space - please consider the following test case:
```scala
val fontAwesomeStyle = CssStyleNameWithSharedCompanion("far", "fa", "fa-grin")
i(fontAwesomeStyle).render shouldBe "<i class=\"far fa-grin\"></i>"
```

css/src/main/scala/io/udash/css/CssText.scala
```diff
@@ -12,6 +12,6 @@ trait CssText {
 object CssText extends CssText {
   private final class TextStyleModifier(styles: CssStyle*) extends Modifier {
```
bgrochal: `CssText#styles2TextMod` (varargs version) is not tested.

Aside question: is there even a possibility to use this implicit conversion?

ddworak: Only explicitly - removing both this one and the one in `CssView`.

PR #390, opened 2019-09-02
Circe upgrade

ddworak:


rest/src/test/scala/io/udash/rest/CirceRestApiTest.scala
```diff
@@ -65,16 +65,16 @@ abstract class HasCirceCodec[T](
...
+  def decoder(nameTransform: String => String, useDefaults: Boolean, discriminator: Option[String]): Decoder[T]
 }
 
 abstract class HasCirceCustomizedCodec[T](nameTransform: String => String)(
   implicit instances: MacroInstances[Unit, CirceCustomizedInstances[T]]
 ) {
-  implicit final lazy val objectEncoder: Encoder.AsObject[T] = instances((), this).encoder(nameTransform)
-  implicit final lazy val decoder: Decoder[T] = instances((), this).decoder(nameTransform)
+  implicit final lazy val objectEncoder: Encoder.AsObject[T] = instances((), this).encoder(nameTransform, None)
+  implicit final lazy val decoder: Decoder[T] = instances((), this).decoder(nameTransform, true, None)
```
ghik: `nameTransform` is accepted in constructor, I believe the other two params should also be (maybe with default values).

PR #405, opened 2019-09-27
Update artifact ID for Bootstrap extensions

ddworak:
The demos are based on the B4 module, whereas the dependency
(and template generator https://github.com/UdashFramework/udash.g8/issues/17)
still mentions the old, B3 based module.


mereszeta:
LGTM

PR #406, opened 2019-09-30
Added RestDataCompanionWithDeps

ghik:


rest/src/main/scala/io/udash/rest/RestDataCompanion.scala
```diff
@@ -32,6 +33,19 @@ abstract class RestDataCompanion[T](implicit
   implicit lazy val restSchema: RestSchema[T] = RestSchema.lazySchema(restStructure.standaloneSchema)
 }
 
+/**
+  * A version of [[RestDataCompanion]] which injects additional implicits into macro materialization.
+  * Implicits are imported from an object specified with type parameter `D`.
+  * It must be a singleton object type, i.e. `SomeObject.type`.
+  */
+abstract class RestDataCompanionWithDeps[D, T](implicit
```
ddworak: Needs a test.

PR #422, opened 2019-10-15
Cookie names and values are URL-encoded

ghik:
This fixes the problem encountered by #410 where Jetty started having problems with cookies containing non-ascii characters. Cookies are notoriously ill-specified and there's no real accepted and implemented standard for them: https://stackoverflow.com/questions/1969232/allowed-characters-in-cookies

This is technically a breaking change but IMHO the chance of this affecting anyone is slim because:
* cookies are rarely used
* cookies with characters which need URL-encoding are even less probable
* projects using Udash REST on both client and server side will be fine as long as both sides are using the same version of Udash REST

### DIFF
--- a/rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
+++ b/rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
@@ -8,6 +8,7 @@ import com.avsystem.commons.annotation.explicitGenerics
 import com.typesafe.scalalogging.LazyLogging
 import io.udash.rest.RestServlet._
 import io.udash.rest.raw._
+import io.udash.utils.URLEncoder
 import javax.servlet.http.{HttpServlet, HttpServletRequest, HttpServletResponse}
 import javax.servlet.{AsyncEvent, AsyncListener}
 
@@ -85,9 +86,13 @@ class RestServlet(
     }
     val headers = headersBuilder.result()
 
-    val cookies = request.getHeaders(CookieHeader).asScala.foldLeft(Mapping.empty[PlainValue]) {
-      (cookies, headerValue) => cookies ++ PlainValue.decodeCookies(headerValue)
+    val cookiesBuilder = Mapping.newBuilder[PlainValue]
+    request.getCookies.opt.getOrElse(Array.empty).foreach { cookie =>
+      val cookieName = URLEncoder.decode(cookie.getName, plusAsSpace = true)
+      val cookieValue = URLEncoder.decode(cookie.getValue, plusAsSpace = true)
+      cookiesBuilder += cookieName -> PlainValue(cookieValue)
     }
+    val cookies = cookiesBuilder.result()
 
     RestParameters(path, headers, query, cookies)
   }
--- a/rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
+++ b/rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
@@ -20,13 +20,13 @@ object RestServlet {
   final val CookieHeader = "Cookie"
 
   /**
-    * Wraps an implementation of some REST API trait into a Java Servlet.
-    *
-    * @param apiImpl        implementation of some REST API trait
-    * @param handleTimeout  maximum time the servlet will wait for results returned by REST API implementation
-    * @param maxPayloadSize maximum acceptable incoming payload size, in bytes;
-    *                       if exceeded, `413 Payload Too Large` response will be sent back
-    */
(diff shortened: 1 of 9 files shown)


rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
```diff
@@ -74,19 +74,21 @@ class RestServlet(
...
       if (!headerName.equalsIgnoreCase(CookieHeader)) { // cookies are separate, don't include them into header params
         headersBuilder += headerName -> PlainValue(request.getHeader(headerName))
       }
     }
     val headers = headersBuilder.result()
-    val cookiesBuilder = Mapping.newBuilder[PlainValue]
-    request.getCookies.opt.getOrElse(Array.empty).foreach { cookie =>
-      cookiesBuilder += cookie.getName -> PlainValue(cookie.getValue)
+
+    val cookies = request.getHeaders(CookieHeader).asScala.foldLeft(Mapping.empty[PlainValue]) {
```
ddworak: Does this mean that we cannot rely on `getCookies` anywhere, anymore? It might be hard to use with filters / interceptors.

rest/src/main/scala/io/udash/rest/SttpRestClient.scala
```diff
@@ -45,17 +45,11 @@ object SttpRestClient {
...
 
-    val cookieHeaders = List(request.parameters.cookies.entries).filter(_.nonEmpty).map { cookies =>
-      "Cookie" -> cookies.iterator.map {
-        case (n, PlainValue(v)) =>
-          require(!v.contains(";"), s"invalid cookie: $n=$v")
-          s"$n=$v"
-      }.mkString(";")
-    }
+    val cookieHeaders = List(request.parameters.cookies).filter(_.nonEmpty)
+      .map(cookies => "Cookie" -> PlainValue.encodeCookies(cookies))
```
ddworak: Should be the same constant as `io.udash.rest.RestServlet#CookieHeader`

rest/src/main/scala/io/udash/rest/raw/PlainValue.scala
```diff
@@ -38,4 +38,19 @@ object PlainValue extends (String => PlainValue) {
...
+  def decodeCookies(cookieHeaderValue: String): Mapping[PlainValue] = {
+    val builder = Mapping.newBuilder[PlainValue]
+    cookieHeaderValue.split(";").iterator.map(_.trim.split("=", 2)).foreach {
+      case Array(name, value) => builder +=
+        URLEncoder.decode(name, plusAsSpace = true) -> PlainValue(URLEncoder.decode(value, plusAsSpace = true))
+      case _ => throw new IllegalArgumentException(s"invalid cookie header value $cookieHeaderValue")
+    }
+    builder.result()
+  }
 }
```
ddworak: Should we update markdown docs somewhere to mention this solution?

ghik:
Your quote does not specify how to escape cookies in general, it just recommends being defensive.

rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
ghik: Any other solution?

ddworak: I'm happy with the way things work, in fact, your implementation is compliant with the Jetty guidelines for this change:

I just want the decoding to be done on the `Cookie` objects as opposed to parsing the whole header, so we can always rely on the way they're composed in other Jetty abstractions. This seems to be possible to do here (simplified code sample):
```scala
val cookies = request.getCookies.opt.getOrElse(Array.empty).foldLeft(Mapping.empty[PlainValue]) {
  (cookies, headerValue) => cookies.append(URLEncoder.decode(headerValue.getName, true), PlainValue(URLEncoder.decode(headerValue.getValue, true)))
}
```

rest/jetty/src/main/scala/io/udash/rest/jetty/JettyRestClient.scala
```diff
@@ -46,7 +47,8 @@ object JettyRestClient {
           case (name, PlainValue(value)) => httpReq.header(name, value)
         }
         request.parameters.cookies.entries.foreach {
-          case (name, PlainValue(value)) => httpReq.cookie(new HttpCookie(name, value))
+          case (name, PlainValue(value)) => httpReq.cookie(new HttpCookie(
+            URLEncoder.encode(name, spaceAsPlus = true), URLEncoder.encode(value, spaceAsPlus = true)))
```
ddworak: Missing test for name parsing.

PR #438, opened 2019-10-25
Added util for conditional require

### LINKED ISSUE #304: AuthView util for conditional require
The task is to add a helper to `io.udash.auth.AuthView` that would enable rendering one template when user has a permission and a different one when he doesn't.

mereszeta:
Added method that allows displaying fallback view when user is unauthorized 

Closes #304


auth/.js/src/main/scala/io/udash/auth/AuthView.scala
```diff
@@ -1,15 +1,27 @@
...
+    if (canRenderView(permission, requireAuthenticated)) view else ()
 
   /** Renders provided `view` only if user is authenticated. */
   def requireAuthenticated(view: => Modifier)(implicit userCtx: UserCtx): Modifier =
-    require(PermissionCombinator.AllowAll, requireAuthenticated = true)(view)
+    require(AllowAll, requireAuthenticated = true)(view)
+
+  /** Renders provided view if user is authenticated or fallback view otherwise */
+  def requireWithFallback(permission: PermissionCombinator, requireAuthenticated: Boolean)
+    (view: => Modifier, fallback: => Modifier)(implicit userCtx: UserCtx): Modifier =
```
mkej: I don't know the most common usage patterns, but with this signature we lock ourselves out of convenient syntax for creating main view, like:
```scala
require(SomePerm) {
  // multi-line code that creates view
}
```

If the `view` and `fallback` would be in separate args list, one could write:
```scala
requireWithFallback(SomePerm) {
  // multi-line code that creates view
}(fallback)
```
or
```scala
requireWithFallback(SomePerm) {
  // multi-line code that creates view
} {
  // multi-line fallback creation
}
```

Of course this is more cosmetic than practical, I'd put it to a vote ;)

mkej: Side note - current signature is way better if views are mostly passed to the function as variable references, not entire blocks of code. That's why I mentioned most common usage patterns.

mereszeta: From what I've seen I believe that the most common usage is this one:
```
require(Permission)(View)
```
The only time I've seen one with block of code was in AuthViewTest. @ddworak what do you think?

ddworak: I have to agree with @mkej, the possibility to use a block here is quite useful (and it's a actually a part of the examples on https://guide.udash.io/ext/authorization, which should be updated within the scope of this PR :smile: )

mereszeta: Understood, I'll adjust the interface and update examples

auth/.js/src/main/scala/io/udash/auth/AuthView.scala
```diff
@@ -1,15 +1,27 @@
 package io.udash.auth
 
+import io.udash.auth.PermissionCombinator.AllowAll
 import scalatags.JsDom.all._
 
 trait AuthView {
   /** Renders provided `view` only if user context has required permissions. */
   def require(permission: PermissionCombinator, requireAuthenticated: Boolean = false)(view: => Modifier)(implicit userCtx: UserCtx): Modifier =
-    if ((!requireAuthenticated || userCtx.isAuthenticated) && permission.check(userCtx)) view else ()
+    if (canRenderView(permission, requireAuthenticated)) view else ()
```
ddworak: Shouldn't this be implemented in terms of `requireWithFallback`?

guide/guide/.js/src/main/scala/io/udash/web/guide/views/ext/AuthorizationExtView.scala
```diff
@@ -82,14 +83,24 @@ class AuthorizationExtView extends FinalView with CssView {
...
+         |      span("This is fallback view when user context is not authenticated")
          |    }
          |  )
          |}
          |""".stripMargin
     )(GuideStyles),
     h2("RPC endpoints authorization"),
     p(
-      "Similar methods are available in ", i("AuthRequires"), " trait, but instead of filtering GUI elements these ",
+      "Similar methods to first two from ", i("AuthView"), " are available in ", i("AuthRequires"), " trait, but instead of filtering GUI elements these ",
```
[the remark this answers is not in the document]
mereszeta: Good idea, I will propose solution in separate pr :smile:

mereszeta:
ping @ddworak

auth/.js/src/main/scala/io/udash/auth/AuthView.scala
```diff
@@ -1,15 +1,25 @@
...
+    requireWithFallback(permission, requireAuthenticated)(view)()
 
   /** Renders provided `view` only if user is authenticated. */
   def requireAuthenticated(view: => Modifier)(implicit userCtx: UserCtx): Modifier =
-    require(PermissionCombinator.AllowAll, requireAuthenticated = true)(view)
+    require(AllowAll, requireAuthenticated = true)(view)
+
+  /** Renders provided `view` if user context has required permissions or `fallback` view otherwise. */
+  def requireWithFallback(permission: PermissionCombinator, requireAuthenticated: Boolean = false)
+    (view: => Modifier)(fallback: => Modifier = ())(implicit userCtx: UserCtx): Modifier =
```
ddworak: You can use `io.udash.bindings.modifiers.EmptyModifier` instead.

PR #444, opened 2019-11-04
Fix routing example in docs

ddworak:


core/.js/src/test/scala/io/udash/routing/WindowUrlPathChangeProviderTest.scala
```diff
@@ -22,8 +22,10 @@ class WindowUrlPathChangeProviderTest extends UdashFrontendTest {
```
bgrochal: No `retrying` necessary above?

PR #449, opened 2019-11-13
Distinguish full and element-by-element transformation

ddworak:


ddworak:
Towards #50

ddworak:
Towards #220

ddworak:
Towards #272

PR #450, opened 2019-11-14
Importless css

ddworak:


ddworak:
This is blocked due to IntelliJ support issues (which are crucial to us).

catap:
@ddworak any news here?

ddworak:
Not much, I'm checking periodically and the IDE support isn't any better.

catap:
@ddworak maybe you have open a ticket for it? I'd love to subscribe :)

ddworak:
I don't think there's anything novel I could submit - cross-projects (neither the way we do them or using `sbt-crossproject`) are not supported by IntelliJ correctly, it cannot distinguish between the JVM and JS shared parts correctly.

catap:
@ddworak I guess we have both the same issue. Let me share with you via gitter's private chat that I've already done to ask them to hunt it down :)

PR #457, opened 2019-11-20
Zipped seq properties should notify about element changes

ddworak:
* registering a listener early should not fail appends in zipped sequences
* .zipAll initial value fix
* PR title should be true as well

### DIFF
--- a/core/src/main/scala/io/udash/properties/seq/ZippedSeqProperty.scala
+++ b/core/src/main/scala/io/udash/properties/seq/ZippedSeqProperty.scala
@@ -12,12 +12,10 @@ private[properties] abstract class ZippedSeqPropertyUtils[O] extends AbstractRea
   protected var children: MBuffer[ReadableProperty[O]] = _
   private val originStructureListener: Patch[ReadableProperty[_]] => Unit =
     (patch: Patch[ReadableProperty[_]]) => {
-      val idx = patch.idx
-      val els = children
-      val removed = CrossCollections.slice(els, patch.idx, els.length)
-      val added = updatedPart(idx)
+      val removed = CrossCollections.slice(children, patch.idx, children.length)
+      val added = updatedPart(patch.idx)
       if (added.nonEmpty || removed.nonEmpty) {
-        CrossCollections.replaceSeq(els, idx, removed.size, added)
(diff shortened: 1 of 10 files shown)


core/src/main/scala/io/udash/properties/seq/ZippedSeqProperty.scala
```diff
@@ -2,42 +2,63 @@ package io.udash.properties.seq
...
+  private final val children = CrossCollections.createArray[ReadableProperty[O]]
+  private final val sourceRegistrations = CrossCollections.createArray[Registration]
+  private final val childrenRegistrations = CrossCollections.createArray[Registration]
+
+  private val originStructureListener: Patch[ReadableProperty[_]] => Unit = { patch =>
+    val removed = CrossCollections.slice(children, patch.idx, children.length)
+    val added = updatedPart(patch.idx)
+    if (added.nonEmpty || removed.nonEmpty) {
+      CrossCollections.replaceSeq(children, patch.idx, removed.size, added)
+      val mappedPatch = Patch(patch.idx, removed.toSeq, added, patch.clearsProperty)
```
najder-k: `.toSeq` unneeded

ddworak: Do you have a moment to talk about Scala 2.13, sir?

core/src/main/scala/io/udash/properties/seq/ZippedSeqProperty.scala
```diff
@@ -2,42 +2,63 @@ package io.udash.properties.seq
...
-      if (added.nonEmpty || removed.nonEmpty) {
-        val mappedPatch = Patch(patch.idx, removed.toSeq, added, patch.clearsProperty)
-        CallbackSequencer().queue(
-          s"${this.id.toString}:fireElementsListeners:${patch.hashCode()}",
-          () => structureListeners.foreach(_.apply(mappedPatch))
-        )
-        valueChanged()
-      }
+private[properties] abstract class ZippedSeqPropertyUtils[O](
+  sources: ISeq[ReadableSeqProperty[_, _ <: ReadableProperty[_]]]
```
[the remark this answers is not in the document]
ddworak: I'll add a comment, there's no need to verify this in runtime.

PR #464, opened 2019-11-22
Improve PropertyCreators usage

ddworak:
* avoid redundant creators - they're only used for creating new direct properties
* cache some creators

### DIFF
--- a/core/.js/src/main/scala/io/udash/Application.scala
+++ b/core/.js/src/main/scala/io/udash/Application.scala
@@ -6,8 +6,6 @@ import io.udash.routing.StateChangeEvent
 import io.udash.utils.CallbacksHandler
 import org.scalajs.dom.Element
 
-import scala.reflect.ClassTag
-
 /**
  * Root application which is used to start single instance of app.
  *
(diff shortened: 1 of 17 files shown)


core/.js/src/main/scala/io/udash/bindings/inputs/Select.scala
```diff
@@ -23,7 +23,7 @@ object Select {
     * @param selectModifiers Additional Modifiers for the select tag, don't use modifiers on value, onchange and selected attributes.
     * @return Binding with `select` element, which can be used as Scalatags modifier.
     */
-  def apply[T : PropertyCreator](
+  def apply[T](
```
bgrochal: Should `UdashForm` methods: `select`, `multiSelect`, `checkButtons` and `radioButtons` take a `PropertyCreator` instance? Same applies to the `ButtonsComponent` class probably.

bgrochal: `PropertyCreator` for `UdashPagination#PageType` seems to be unnecessary - please check.

core/.js/src/main/scala/io/udash/Application.scala
```diff
@@ -15,7 +15,7 @@ import scala.reflect.ClassTag
  * @param viewFactoryRegistry [[io.udash.core.ViewFactoryRegistry]] implementation, which will be used to match [[io.udash.core.State]] into [[io.udash.core.ViewFactory]]
  * @tparam HierarchyRoot Should be a sealed trait which extends [[io.udash.core.State]].
  */
-class Application[HierarchyRoot >: Null <: GState[HierarchyRoot] : ClassTag : PropertyCreator](
+class Application[HierarchyRoot >: Null <: GState[HierarchyRoot] : PropertyCreator](
```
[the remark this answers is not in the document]
ddworak: If we don't and the state has a model property creator, we might end up creating a regular, direct property and have runtime `.asModel` failures.

PR #465, opened 2019-11-22
Remove udash-charts

ddworak:
We're no longer supporting udash-charts after switching internally to https://github.com/oyvindberg/ScalablyTyped/tree/master/h/highcharts, which is automatically updated, have all new features and not that much worse experience.


mowczare:
LGTM

PR #467, opened 2019-11-28
View#onClose

ddworak:
This should simplify integration with component frameworks which manage bindings internally.

### DIFF
--- a/core/.js/src/test/scala/io/udash/routing/RoutingEngineTest.scala
+++ b/core/.js/src/test/scala/io/udash/routing/RoutingEngineTest.scala
@@ -25,8 +25,9 @@ class RoutingEngineTest extends UdashFrontendTest with TestRouting {
         ErrorState -> errorViewFactory
       )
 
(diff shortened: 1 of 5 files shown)


core/.js/src/test/scala/io/udash/routing/RoutingEngineTest.scala
```diff
@@ -2,106 +2,140 @@ package io.udash.routing
 
 import io.udash._
 import io.udash.testing._
+import org.scalactic.source.Position
 
 class RoutingEngineTest extends UdashFrontendTest with TestRouting {
 
   "RoutingEngine" should {
     "render valid views on url change" in {
```
mereszeta: Could we split this test case? It's really long and description is very general.

mereszeta:
LGTM besides one question

core/.js/src/test/scala/io/udash/routing/RoutingEngineTest.scala
ddworak: I considered it briefly, but it led to another test case of similar length - without a few assertions. I'll keep this as integration test for now.

Issue #396, opened 2019-09-10
ScalaFiddle demos update


ddworak:

ddworak:
First step: https://github.com/scalafiddle/scalafiddle-io/pull/41

PR #480, opened 2019-12-12
Fiddle

ddworak:


ddworak:
Closes #396

ddworak:
@bgrochal I see your point from a "production code" standpoint, but I'm not sure they'll make a homepage intro demo better.

- `a10A6UA` - we could just `bind`, but then it doesn't only append DOM nodes for new numbers, it rerenders the whole line, which doesn't showcase the property mechanism well enough. You could argue to use `repeatWithIndex`, but `.get`ting here is actually optimal, as long as we are guaranteed to only have structure changes.

- `ButwLWQ` - this input is pre-filled and the input doesn't require any additional confirmation (e.g. Enter press), so the placeholder is omitted to make the demo simpler. `errors` is just to inform the reader about the content here, there's no IntelliJ at hand on udash.io ;)

PR #486, opened 2019-12-16
Fix docker run for guide

### LINKED ISSUE #409: Enable docker packaging for guide
Currently, the Udash Guide can only be packaged via:
```
sbt guide-packager/universal:packageBin
```
We'd like to add Docker packaging based on SBT Native Packager [Docker plugin](https://www.scala-sbt.org/sbt-native-packager/formats/docker.html), so one could e.g.:
```
sbt guide-packager/docker:publishLocal
```

`docker run` should start guide without errors by default.

ddworak:
This should:
- remove warnings on `sbt guide-packager/docker:publishLocal`
- fix file logging issue in docker (by disabling it conditionally)
- not break file logging in deb

Closes #409

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -470,6 +470,7 @@ lazy val `guide-packager` =
       },
 
       dockerExposedPorts += 8080,
+      dockerEnvVars += "DISABLE_FILE_LOGGING" -> "true",
     )
 
 lazy val `guide-selenium` =
--- a/build.sbt
+++ b/build.sbt
@@ -468,6 +468,8 @@ lazy val `guide-packager` =
         val frontendStatics = (`guide-guide` / Compile / compileAndOptimizeStatics).value
         (frontendStatics.allPaths --- frontendStatics) pair relativeTo(frontendStatics.getParentFile)
       },
+
+      dockerExposedPorts += 8080,
     )
 
 lazy val `guide-selenium` =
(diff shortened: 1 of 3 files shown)


project/Dependencies.scala
```diff
@@ -183,6 +184,7 @@ object Dependencies {
   val backendDeps = Def.setting(Seq(
     "com.typesafe.scala-logging" %% "scala-logging" % scalaLoggingVersion,
     "ch.qos.logback" % "logback-classic" % logbackVersion,
+    "org.codehaus.janino" % "janino" % janinoVersion,
```
mowczare: Can't really tell where it is used, could you clarify?

build.sbt
```diff
@@ -468,6 +468,9 @@ lazy val `guide-packager` =
         val frontendStatics = (`guide-guide` / Compile / compileAndOptimizeStatics).value
         (frontendStatics.allPaths --- frontendStatics) pair relativeTo(frontendStatics.getParentFile)
       },
+
+      dockerExposedPorts += 8080,
```
mowczare: NIT: you may add comment here indicating this port should match ui server port.

project/Dependencies.scala
ddworak: Welcome to the Java world, where library capabilities are determined on runtime based on the JARs that are loaded. Basically conditional statements in logback require this - adding a comment.

PR #493, opened 2019-12-22
Simplified Udash Listenable

ddworak:


core/.js/src/main/scala/io/udash/component/Listenable.scala
```diff
@@ -1,27 +1,45 @@
...
-  private val actions = new CallbacksHandler[EventType]
+trait Listenable {
+  type EventType <: ListenableEvent
+  final type EventHandler = PartialFunction[EventType, Unit]
+
+  //visible for testing
+  private[udash] val listeners = MLinkedHashSet.empty[EventHandler]
+
+  /**
+   * Register event handler in component.
```
bgrochal: Nit: I would recommend to improve API docs slightly:
1. _Registers_ and _Removes_ (line 25) should be probably better,
2. `Registration` object allows to do more with a listener than just to deregister it (line 17),
3. we could add some sort of docs to the `Listenable` and `ListenableEvent` traits, considering callbacks execution order and other general aspects.

bootstrap4/.js/src/main/scala/io/udash/bootstrap/tooltip/TooltipUtils.scala
```diff
@@ -13,7 +13,10 @@ import scala.concurrent.duration.{Duration, DurationInt}
 import scala.scalajs.js
 import scala.scalajs.js.|
 
-trait Tooltip[EventType <: ListenableEvent[ThisType], ThisType <: Tooltip[EventType, ThisType]] extends Listenable[ThisType, EventType] {
+trait Tooltip extends Listenable {
+
+  override type EventType = TooltipEvent
```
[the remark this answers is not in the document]
ddworak: It's pretty hard to override a type badly once it's assigned, but sure, why not.

core/.js/src/test/scala/io/udash/component/ListenableTest.scala
```diff
@@ -0,0 +1,123 @@
...
+      case _ =>
+    }
+  }
+
+  private trait Fixture {
+    protected val listenable = new TestListenable
+  }
+
+  "Listenable" should {
+    "create an active Registration on listen called" in new Fixture {
```
bgrochal: nit: this test case is semantically related to the "Registration" suite IMO

core/.js/src/test/scala/io/udash/component/ListenableTest.scala
```diff
@@ -0,0 +1,123 @@
...
+
+  def fireEvent(): Unit = {
+    fire(TestListenableEvent(this))
+  }
+}
+private final case class TestListenableEvent(source: TestListenable) extends ListenableEvent
+
+class ListenableTest extends UdashFrontendTest {
+
+  def getEmptyRegistration(listenable: Listenable): Registration = {
```
bgrochal: nit: could be a part of `TestListenable` implementation since `listenable` is always passed as the one and only argument

PR #494, opened 2019-12-22
Introduced support of tags inside translation.

catap:
Right now translation ignore all HTML tags and insert them as text and make something like: `some <b>bold</b> text` is very tricky: you should concat it inside view.

This commit introduces a native support for this sort of string.

### DIFF
--- a/i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationBinding.scala
+++ b/i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationBinding.scala
@@ -2,7 +2,8 @@ package io.udash.i18n.bindings
 
 import com.avsystem.commons._
 import io.udash.bindings.Bindings
-import io.udash.i18n.Translated
+import io.udash.bindings.modifiers._
+import io.udash.i18n._
 import io.udash.logging.CrossLogging
 import org.scalajs.dom._
 import scalatags.JsDom.Modifier
@@ -10,24 +11,19 @@ import scalatags.JsDom.Modifier
 import scala.concurrent.Future
 import scala.util.{Failure, Success}
 
-private[i18n] class TranslationBinding(translation: Future[Translated], placeholder: Option[Element])
-  extends Modifier with Bindings with CrossLogging {
+private[i18n] final class TranslationBinding(
+  translation: Future[Translated],
+  placeholder: Option[Element],
+  rawHtml: Boolean
+) extends Modifier with Bindings with CrossLogging {
 
-  override def applyTo(t: Element): Unit =
-    if (translation.isCompleted && translation.value.get.isSuccess) {
-      t.appendChild(document.createTextNode(translation.value.get.get.string))
-    } else {
-      val holder: Node = placeholder.getOrElse(emptyStringNode())
-      t.appendChild(holder)
-
-      translation.onCompleteNow {
-        case Success(text) =>
-          t.replaceChild(
-            document.createTextNode(text.string),
-            holder
-          )
-        case Failure(ex) =>
-          logger.error(ex.getMessage)
-      }
+  override def applyTo(t: Element): Unit = {
+    val holder: Seq[Node] = t.appendChild(placeholder.getOrElse(emptyStringNode()))
+    translation onCompleteNow {
+      case Success(Translated(text)) =>
+        t.replaceChildren(holder, parseTranslation(rawHtml, text))
+      case Failure(ex) =>
+        logger.error(ex.getMessage)
     }
+  }
 }
--- a/i18n/.js/src/main/scala/io/udash/i18n/bindings/package.scala
+++ b/i18n/.js/src/main/scala/io/udash/i18n/bindings/package.scala
@@ -1,0 +1,20 @@
+package io.udash.i18n
+
+import org.scalajs.dom._
+
+import scala.scalajs.js.{JavaScriptException, SyntaxError}
+import scala.util.{Failure, Success, Try}
+
+package object bindings {
+  def parseTranslation(rawHtml: Boolean, text: String): Seq[Node] =
+    if (rawHtml) Try {
+      val wrapper = document.createElement("div")
+      wrapper.innerHTML = text
+      wrapper.childNodes
+    } match {
+      case Success(children) if children.length > 0 =>
+        (0 until children.length).map(children.item)
+      case Success(_) | Failure(JavaScriptException(_: SyntaxError)) =>
+        Seq(document.createTextNode(text))
+    } else Seq(document.createTextNode(text))
+}
--- a/i18n/.js/src/test/scala/io/udash/i18n/BindingsTest.scala
+++ b/i18n/.js/src/test/scala/io/udash/i18n/BindingsTest.scala
@@ -10,21 +10,24 @@ class BindingsTest extends AsyncUdashFrontendTest {
 
   implicit val provider: TranslationProvider = new LocalTranslationProvider(Map(
     Lang("en") -> Bundle(BundleHash("hash1"), Map(
+      "tr0" -> null,
       "tr1" -> "Translation {0}",
       "tr2" -> "Translation2 {1} {0}",
       "tr3" -> "Translation3 {}",
-      "tr4" -> "Translation4 {1} {} {}",
-      "tr5" -> "Translation5 {4}"
+      "tr4" -> "Translation4 <b>{1}</b> {} {}",
+      "tr5" -> "Translation5 <b>{4}</b>"
     )),
     Lang("pl") -> Bundle(BundleHash("hash1"), Map(
+      "tr0" -> null,
       "tr1" -> "Translation {0} pl",
       "tr2" -> "Translation2 {1} {0} pl",
       "tr3" -> "Translation3 {} pl",
-      "tr4" -> "Translation4 {1} {} {} pl",
-      "tr5" -> "Translation5 {4} pl"
+      "tr4" -> "Translation4 <b>{1}</b> {} {} pl",
+      "tr5" -> "Translation5 <b>{4}</b> pl"
     ))
   ), missingTranslationError = "ERROR")
 
+  val key0 = TranslationKey.keyX("tr0")
   val key1 = TranslationKey.key1[String]("tr1")
   val key2 = TranslationKey.key2[Int, Double]("tr2")
   val key3 = TranslationKey.key1[Double]("tr3")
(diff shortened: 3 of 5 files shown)


ddworak:
@catap thanks for the contribution! :tada: We'll need some time to consider whether this should be the default behaviour, since there are quite a few issues regarding:
- compatibility with Java i18n / resource bundles - not sure if we need to maintain it, but should these translations be shared between Udash and another app or translated on the backend, there would be differences in the resulting strings.
- backward compatibility - some tags included in user translations might be parsed now if not escaped properly and that's pretty hard to detect automatically.
- performance - parsing all text may be slower (to be benchmarked).
- flexibility vs. security - what about `<script>` or non-inline HTML tags? It's not the most exciting attack vector, but I certainly hope to prevent abusing the translation engine for templating.

I'll get back to you as soon as I have more input on these.

catap:
@ddworak well

1) I love idea to keep all strings for backend and frontend at one place and resource bundle looks good for me for that.

I do have a bit more code, because I've implemented a way to change switch backend locale by `Accept-Header`.

I will create a few more pull requests to i18n part to merge this changes soon.

2) I thought about that.

Because you pointed it and it is good point, I will make force push ASAP that make this behaviour is optional and switched off by default compromise I can switched to flag. Something like
```
  def translatedDynamic[Key <: TranslationKey](
    key: Key, placeholder: Option[Element] = None, rawHtml = false
  )(
    translator: Key => Future[Translated]
  )(implicit lang: LangProperty): Binding =
    new DynamicTranslationBinding(key, translator, placeholder, rawHtml)
```

That allows to enable it where is needed

3) Agreed, and (2) should mask this issue

4) Well, if we started to discuss about security I see only one way how to exploit this: put somethings bad inside translation cache.

At my project I've "switched off" cache, and just fetch all translation by one request.

But it is bad design for now, better design is move translation into shared code base that is fine, but scala js doesn't support ResourceBundle and Locale :) that is expected, and was a blocker for me to do it.

I'm thinking a bit how to do it on the right way, but I feel that it is out of scope this pull request.

Rigth now I'm building complex system with udash that should have i18n at backend and frontend, I'd like to contribute all changes that I did with i18n if it fine for you ;)

i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationBinding.scala
```diff
@@ -5,29 +5,42 @@ import io.udash.bindings.Bindings
...
           logger.error(ex.getMessage)
       }
-    }
+  }
+
+  override def applyTo(t: Element): Unit =
+    impl(placeholder, t.appendChild, t.replaceChild)
+}
+
+private[i18n] class RawTranslationBinding(
```
ddworak: nit: `final`

i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationBindingHelper.scala
```diff
@@ -0,0 +1,48 @@
+package io.udash.i18n.bindings
+
+import io.udash.logging.CrossLogging
+import org.scalajs.dom._
+import org.scalajs.dom.raw.Node
+
+private[bindings] trait TranslationBindingHelper extends CrossLogging {
+  import TranslationBindingHelper._
+
+  implicit def String2Node(str: String): Node =
```
ddworak: Implicit conversions on very commonly used types are considered a bad practice and tend to break when least expected. Especially in this case, since you're shadowing the default scalatags conversions for rendering strings.

i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationBindingHelper.scala
```diff
@@ -0,0 +1,48 @@
...
+      kids.length match {
+        case l if l > 0 =>
+          (0 until kids.length).map(kids(_))
+        case _ =>
+          logger.warn(s"'$str' is parsed as nothing, use it as string.")
+          val node: Node = document.createTextNode(str)
+          Seq(node)
+      }
+    } catch {
+      case e: Throwable =>
```
ddworak: Can we narrow this to a more specific class of exceptions? This may very well have been an exception from `document.createTextNode(str)` and fail again.

i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationBindingHelper.scala
```diff
@@ -0,0 +1,48 @@
...
+        logger.warn(s"Can't parse: '$str', use it as string. Message: ${e.getMessage}")
+        val node: Node = document.createTextNode(str)
+        Seq(node)
```
ddworak: I think this error handling is common with the empty node list case.

i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationBindingHelper.scala
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+  implicit def Element2ExtendedElement(e: Element): ExtendedElement =
```
ddworak: nit: method names are commonly lower camel cased

i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationBindingHelper.scala
```diff
@@ -0,0 +1,48 @@
...
+import org.scalajs.dom.raw.Node
+
+private[bindings] trait TranslationBindingHelper extends CrossLogging {
+  import TranslationBindingHelper._
+
+  implicit def String2Node(str: String): Node =
+    document.createTextNode(str)
+
+  implicit def String2Nodes(str: String): Seq[Node] =
+    try {
```
ddworak: You can consider reusing the `raw` tag from scalatags to implement this:  `https://www.lihaoyi.com/scalatags/#Auto-escapingandunsanitizedInput`

i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationBinding.scala
```diff
@@ -5,29 +5,42 @@ import io.udash.bindings.Bindings
...
+  translation: Future[Translated],
+  placeholder: Option[Element]
+) extends TranslationBinding(translation, placeholder) {
+  override def applyTo(t: Element): Unit =
+    impl(placeholder.map(Seq(_)), t.appendKids, t.replaceKids)
```
ddworak: You might want to use `io.udash.bindings.modifiers.ElementExts#replaceChildren` for multiple nodes replacement and inline appending. It should work in both cases if we assume `impl` works on `Seq[Node]`.

i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationBinding.scala
catap: Thanks, fixed

catap:
@ddworak thanks for review. I've pushed a commit that contains part of this fixes.

Soon I will response to another part of review. After that I'll made `rebase -i` for this commits to squash it into single commit.

i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationBinding.scala
```diff
@@ -5,29 +5,42 @@ import io.udash.bindings.Bindings
...
+) extends Modifier with Bindings with CrossLogging with TranslationBindingHelper {
 
-  override def applyTo(t: Element): Unit =
-    if (translation.isCompleted && translation.value.get.isSuccess) {
-      t.appendChild(document.createTextNode(translation.value.get.get.string))
-    } else {
-      val holder: Node = placeholder.getOrElse(emptyStringNode())
-      t.appendChild(holder)
+  protected def impl[T](placeholder: Option[T], append: T => T, replace: (T, T) => T)(
+    implicit str2T: String => T, node2T: Node => T
```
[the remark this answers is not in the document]
catap: Definitely. Very good point.

i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationBinding.scala
catap: Good idea, thanks!

i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationBindingHelper.scala
catap: It was original idea to be honest but I can't do it. Ok, I can but it requires a lot of code, because `raw` way requires to build up a full frag.

catap:
Well, I've fixed pointed issued, made code simpler, and rebased to last master also.

i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationBinding.scala
```diff
@@ -2,32 +2,33 @@ package io.udash.i18n.bindings
...
-    } else {
-      val holder: Node = placeholder.getOrElse(emptyStringNode())
-      t.appendChild(holder)
+  override def applyTo(t: Element): Unit = translation.value match {
+    case Some(Success(Translated(text))) =>
+      parseTranslation(rawHtml, text).foreach(t.appendChild)
 
+    case _ =>
+      val holder: Seq[Node] = t.appendChild(placeholder.getOrElse(emptyStringNode()))
+      holder.foreach(t.appendChild)
```
ddworak: I think you don't need to append it again - you already do in the line above.

i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationBinding.scala
```diff
@@ -2,32 +2,33 @@ package io.udash.i18n.bindings
...
+) extends Modifier with Bindings with CrossLogging {
 
-  override def applyTo(t: Element): Unit =
-    if (translation.isCompleted && translation.value.get.isSuccess) {
-      t.appendChild(document.createTextNode(translation.value.get.get.string))
-    } else {
-      val holder: Node = placeholder.getOrElse(emptyStringNode())
-      t.appendChild(holder)
+  override def applyTo(t: Element): Unit = translation.value match {
+    case Some(Success(Translated(text))) =>
```
ddworak: I'd opt for removing this questionable optimization. With `onCompleteNow` it should be almost the same and the `holder` is just a reference here, it shouldn't prevent GC after the callback is executed.

i18n/.js/src/main/scala/io/udash/i18n/bindings/package.scala
```diff
@@ -0,0 +1,20 @@
...
+import scala.util.{Failure, Success, Try}
+
+package object bindings {
+  def parseTranslation(rawHtml: Boolean, text: String): Seq[Node] =
+    if (rawHtml) Try {
+      val wrapper = document.createElement("div")
+      wrapper.innerHTML = text
+      wrapper.childNodes
+    } match {
+      case Success(kids) if kids.length > 0 =>
```
ddworak: nit: `children`

i18n/.js/src/test/scala/io/udash/i18n/BindingsTest.scala
```diff
@@ -158,22 +160,26 @@ class BindingsTest extends AsyncUdashFrontendTest {
           translatedDynamic(key2)(key => key(3, 3.14)),
           translatedDynamic(key3)(key => key(0.99)),
           translatedDynamic(key4)(key => key("test", 1, true, 3.1415)),
-          translatedDynamic(keyX)(key => key("test", 1.asInstanceOf[Any], true, 3.1415, "Udash"))
+          translatedDynamic(keyX, rawHtml = true)(key => key("test", 1.asInstanceOf[Any], true, 3.1415, "Udash"))
```
ddworak: How about a few more test cases for the raw binding? At least the success with 0 children case is tricky.

i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationBinding.scala
catap: Yep, it was a wrong copy and paste

i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationBinding.scala
catap: @ddworak I don't think that `onCompleteNow` here will be same when you have latency with backend.

Probably I do not understand how `onCompleteNow` works, but for me it executed as soon as future is complete (aka response from backend is here), and if we have latency, we will see placehoder instead nothing.

Have I missed something?

i18n/.js/src/test/scala/io/udash/i18n/BindingsTest.scala
catap: @ddworak good idea. `key0` makes this trick.

i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationBinding.scala
catap: Ok, I guess I've got your idea about `onCompleteNow`. Updated this.

PR #504, opened 2020-01-13
Callback sequencer revamp

ddworak:
Changed introduced due to multiply combined properties firing twice with the same value (see tests). Also makes `.queue` have single semantics regardless of sequencer state.

### DIFF
--- a/core/.js/src/main/scala/io/udash/bindings/inputs/RangeInput.scala
+++ b/core/.js/src/main/scala/io/udash/bindings/inputs/RangeInput.scala
@@ -38,7 +38,7 @@ object RangeInput {
       propertyListeners += valueStep.listen({ v =>
         (step := v).applyTo(element)
         property.set(element.valueAsNumber)
-      }, true)
+      }, initUpdate = true)
 
       override def render: JSInput = element
     }
--- a/core/.js/src/main/scala/io/udash/bindings/inputs/RangeInput.scala
+++ b/core/.js/src/main/scala/io/udash/bindings/inputs/RangeInput.scala
@@ -27,18 +27,18 @@ object RangeInput {
 
       element.onchange = (_: Event) => property.set(element.valueAsNumber)
       propertyListeners += property.listen(element.valueAsNumber = _, initUpdate = true)
-      propertyListeners += minValue.listen { v =>
+      propertyListeners += minValue.listen({ v =>
         (min := v).applyTo(element)
         property.set(element.valueAsNumber)
-      }
-      propertyListeners += maxValue.listen { v =>
+      }, initUpdate = true)
+      propertyListeners += maxValue.listen({ v =>
         (max := v).applyTo(element)
         property.set(element.valueAsNumber)
-      }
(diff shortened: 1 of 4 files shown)


core/.js/src/main/scala/io/udash/bindings/inputs/RangeInput.scala
```diff
@@ -23,18 +23,22 @@ object RangeInput {
...
+        property.set(element.valueAsNumber)
+      }, initUpdate = true)
+      propertyListeners += maxValue.listen({ v =>
+        (max := v).applyTo(element)
+        property.set(element.valueAsNumber)
+      }, initUpdate = true)
+      propertyListeners += valueStep.listen({ v =>
+        (step := v).applyTo(element)
+        property.set(element.valueAsNumber)
+      }, true)
```
bgrochal: nit: missing `Boolean` argument name

core/src/test/scala/io/udash/properties/PropertyTest.scala
```diff
@@ -560,6 +560,41 @@ class PropertyTest extends UdashCoreTest {
...
+      val p2 = Property.blank[Int]
+
+      p0.combine(p1)((_, _))
+        .combine(p2)((_, _))
+        .listen { case ((i1, i2), i3) =>
+          calls.append((i1, i2, i3))
+        }
+
+      CallbackSequencer().sequence {
+        p0.set(0)
```
bgrochal: `p0` default value is `0` (resulting from corresponding `Blank`), hence this line has no effect which is confusing a little bit when digging into the `CallbackSequencer` details - please turn the `force` flag on or change the value

core/src/test/scala/io/udash/properties/PropertyTest.scala
```diff
@@ -560,6 +560,41 @@ class PropertyTest extends UdashCoreTest {
...
+      val p1 = Property.blank[Int]
+      val p2 = Property.blank[Int]
+
+      p0.combine(p1)((_, _))
+        .combine(p2)((_, _))
+        .listen { case ((i1, i2), i3) =>
+          calls.append((i1, i2, i3))
+        }
+
+      CallbackSequencer().sequence {
```
bgrochal: Please add a test case for the `(p2, p1, p0)` setting order - it is probably the most complicated one, which requires passing through the sequencer's `queue` three times.

core/src/test/scala/io/udash/properties/PropertyTest.scala
```diff
@@ -560,6 +560,41 @@ class PropertyTest extends UdashCoreTest {
...
+      calls should contain inOrderElementsOf Seq((0, 1, 2))
+    }
+
+    "short-circuit loops on self" in {
+      val p0 = Property.blank[Int]
+      p0.listen { v => if (v < 10) p0.set(v + 1) }
+
+      p0.set(1)
+
+      p0.get shouldBe 2
```
bgrochal: Should corresponding callback be invoked once outside the `CallbackSequencer`?

Issue #401, opened 2019-09-16
UdashDatePicker in dynamic views


Starzu:
It seems that `UdashDatePicker` does not use provided options when it is a part of a dynamic view. In the following example the date picker works good (it updates the property with selected date), but it does not use provided options. When the date picker is a part of the initial tab it also works ok. 

```scala
def createDatePicker(date: Property[Timestamp]): UdashDatePicker = {
  val options: UdashDatePicker.DatePickerOptions = new UdashDatePicker.DatePickerOptions(
    format = "DD MMMM YYYY", locale = Some("pl")
  )
  UdashDatePicker(date.transform(d => Option(d.toJDate), opt => opt.map(_.toTimestamp).getOrElse(Timestamp.now())), Property(options))()
}

private val tabContent = {
  div(
    UdashInputGroup()(
      UdashInputGroup.prependText("Wybierz dzień: "),
      UdashInputGroup.input(
        createDatePicker(model.subProp(_.eventsDate)).render
      ),
    ).render,
    ???
  )
}

div(
  produce(selectedTab) {
    case Tab1 => ???
    case Tab2 =>tabContent.render
  }
)
```

bgrochal:
@Starzu, could you please provide a more exhaustive snippet and/or detailed reproduction steps for this bug? Unfortunately, I cannot reproduce the issue.

Here's what I've tried:
```scala
object SandboxViewFactory extends StaticViewFactory[SandboxState.type](() => new SandboxView(
  ModelProperty(Model("test", new JDate(1546297200000L).toTimestamp))
))

class SandboxView(model: ModelProperty[Model]) extends FinalView with CssView with JBasicUtils {
  def createDatePicker(date: Property[Timestamp]): UdashDatePicker = {
    val options: UdashDatePicker.DatePickerOptions = new UdashDatePicker.DatePickerOptions(
      format = "DD MMMM YYYY", locale = Some("pl")
    )
    UdashDatePicker(date.transform(d => Option(d.toJDate), opt => opt.map(_.toTimestamp).getOrElse(Timestamp.now())), Property(options))()
  }

  private val tabContent = {
    div(
      UdashDatePicker.loadBootstrapDatePickerStyles(),
      UdashInputGroup()(
        UdashInputGroup.prependText("Wybierz dzień: "),
        UdashInputGroup.input(
          createDatePicker(model.subProp(_.eventsDate)).render
        ),
      ).render,
    )
  }

[...]

PR #512, opened 2020-01-21
Separate DatePicker initialization from options propagation

bgrochal:
When a TempusDominus DatePicker is embedded in a dynamic view (a popup/modal for instance), its options are not initialized properly when passed as a dictionary. The solution is to initialize the picker at first, and to set options sequentially then.

It should not affect picker's performance.

Might be related to #401.

More on options setup in [related docs](https://tempusdominus.github.io/bootstrap-4/Options/).


mereszeta:
Api is not changed in ```UdashDatePickerTest.scala```. I believe that necessary change should be introduced and tests revisited to decide whether we can cover this change via unit tests.

PR #514, opened 2020-01-23
Prevent rerendering of static views by default

ddworak:


core/.js/src/main/scala/io/udash/core/Defaults.scala
```diff
@@ -1,9 +1,17 @@
...
+ * Creates view with [[io.udash.core.EmptyPresenter]]. Used for static views.
+ *
+ * By default, instances of this class are compared by class name to prevent rerendering of static views.
+ * This behaviour can be opted out of by overriding equals/hashCode.
+ **/
+abstract class StaticViewFactory[S <: State](viewCreator: () => View) extends ViewFactory[S] {
   override def create(): (View, EmptyPresenter.type) =
     (viewCreator(), EmptyPresenter)
+
+  override def equals(other: Any): Boolean = getClass.equals(other.getClass)
```
bgrochal: That's a rather unlikely case, but `other.getClass` may throw a NullPointerException here.

core/.js/src/test/scala/io/udash/routing/RoutingEngineTest.scala
```diff
@@ -37,7 +38,7 @@ class RoutingEngineTest extends UdashFrontendTest with TestRouting {
         }
       }
 
-      initTestRoutingEngine(state2vp = state2VP)
+      initTestRoutingEngine(state2vp = state2VP.view.mapValues(() => _).toMap)
```
bgrochal: Nit (applies to the line 296 too): is using `view` and `toMap` necessary if there is only one operation to be executed on map values?

core/.js/src/test/scala/io/udash/routing/RoutingEngineTest.scala
```diff
@@ -327,5 +328,44 @@ class RoutingEngineTest extends UdashFrontendTest with TestRouting {
...
+    "not rerender static views" in {
+      val staticView = new TestView
+      var staticCreateCount = 0
+      def staticViewFactory() = new StaticViewFactory[RootState](() => staticView) {
+        override def create(): (View, EmptyPresenter.type) = {
+          staticCreateCount += 1
+          super.create()
+        }
+      }
+      val state2VP: Map[TestState, () => ViewFactory[_ <: TestState]] = Map[TestState, () => ViewFactory[_ <: TestState]](
```
bgrochal: Nit (applies to the line 19 too): one of explicit type declarations is unnecessary.

core/.js/src/test/scala/io/udash/routing/RoutingEngineTest.scala
ddworak: No `mapValues` in Scala 2.13

PR #525, opened 2020-02-10
Initial index values for `repeatWithIndex` should not be 0.

bgrochal:


core/.js/src/main/scala/io/udash/bindings/modifiers/SeqPropertyWithIndexModifier.scala
```diff
@@ -18,7 +18,7 @@ class SeqPropertyWithIndexModifier[T, E <: ReadableProperty[T]](override val pro
 
   protected def indexProperty(p: E): ReadableProperty[Int] =
     if (indexes.contains(p)) indexes(p)
-    else Property(0).setup { indexes(p) = _ }
+    else Property(property.elemProperties.indexOf(p).applyIf(_ == -1)(_ => 0)).setup(indexes(p) = _)
```
ddworak: `.getOrElseUpdate` can make this method more readable (without if and setup)

PR #526, opened 2020-02-10
DocumentFragments should be rendered conditionally

bgrochal:
- DOM Nodes should be passed to `showIf` and `showIfElse` methods as call-by-name parameters to assure consistency with `produce` and `repeat`, as well as to enable conditional rendering of mutable nodes (like `DocumentFragment`)
- `produce` API unified (no backwards compatibility violation)

### DIFF
--- a/core/.js/src/main/scala/io/udash/bindings/Bindings.scala
+++ b/core/.js/src/main/scala/io/udash/bindings/Bindings.scala
@@ -118,7 +118,7 @@ trait Bindings {
     * @return property binding.
     */
   def produce[T](property: ReadableProperty[T], checkNull: Boolean = true)(builder: T => Seq[Node]): Binding =
-    new PropertyModifier[T](property, builder, checkNull, DOMManipulator.DefaultElementReplace)
+    produce(property, DOMManipulator.DefaultElementReplace, checkNull)(builder)
 
   /**
    * Use it to bind property into DOM structure, given `builder` will be used to generate DOM element on every value change.
@@ -159,7 +159,7 @@ trait Bindings {
     */
   def produceWithNested[T](property: ReadableProperty[T], checkNull: Boolean = true)
     (builder: (T, Binding.NestedInterceptor) => Seq[Node]): Binding =
-    new PropertyModifier[T](property, builder, checkNull, DOMManipulator.DefaultElementReplace)
+    produceWithNested(property, DOMManipulator.DefaultElementReplace, checkNull)(builder)
 
   /**
     * Use it to bind property into DOM structure, given `builder` will be used to generate DOM element on every value change.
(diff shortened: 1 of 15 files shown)


core/.js/src/main/scala/io/udash/bindings/Bindings.scala
```diff
@@ -122,7 +118,24 @@ trait Bindings {
     * @return property binding.
     */
   def produce[T](property: ReadableProperty[T], checkNull: Boolean = true)(builder: T => Seq[Node]): Binding =
-    new PropertyModifier[T](property, builder, checkNull)
+    new PropertyModifier[T](property, builder, checkNull, DOMManipulator.DefaultElementReplace)
```
ddworak: You can use the variant you introduced here (same for `produceWithNested`)

core/.js/src/main/scala/io/udash/bindings/modifiers/PropertyModifier.scala
```diff
@@ -11,21 +11,11 @@ private[bindings] class PropertyModifier[T](
   override val customElementsReplace: DOMManipulator.ReplaceMethod
 ) extends ValueModifier[T] {
 
-  def this(property: ReadableProperty[T], builder: T => Seq[Node], checkNull: Boolean) = {
-    this(property, (data: T, _: Binding.NestedInterceptor) => builder(data), checkNull, DOMManipulator.DefaultElementReplace)
+  def this(property: ReadableProperty[T], builder: T => Seq[Node], checkNull: Boolean, customElementReplace: DOMManipulator.ReplaceMethod) = {
```
ddworak: It's `Elements` in bindings and `Element` here - let's use one (probably the one from bindings as it has way more external usage).

core/.js/src/test/scala/io/udash/bindings/TagsBindingTest.scala
```diff
@@ -1040,12 +1070,12 @@ class TagsBindingTest extends UdashFrontendTest with Bindings { bindings: Bindin
...
       val template = div(
         span(),
-        produceWithNested(p, customElementsReplace = customReplace(true))((v, _) => div(v.mkString(",")).render),
+        produce(p, customElementsReplace = customReplace(true))(v => div(v.mkString(",")).render),
         span()
       ).render
       val template2 = div(
         span(),
-        produceWithNested(p, customElementsReplace = customReplace(false))((v, _) => div(v.mkString(",")).render),
+        produce(p, customElementsReplace = customReplace(false))(v => div(v.mkString(",")).render),
```
ddworak: So, `produceWithNested` can now skip the argument? Improve coverage, not maintain it.

PR #538, opened 2020-02-24
Fix BS4 DatePicker initialization with pre-populated input and jQuery callbacks cleanup

bgrochal:
1. Since custom jQuery event listeners are set each time a DatePicker gets appended to the DOM, there might be multiple listeners attached to the picker embedded within a dynamic view (pic). The fix is to deregister such listeners when detaching the picker from the DOM (using an already configured `MutationObserver` instance).

image

2. DatePicker initialization (separated from options propagation within #512) causes:
```
frontend-deps.js:2 Uncaught RangeError: Maximum call stack size exceeded
    at String.replace (<anonymous>)
    ...
    at w.fn.init.trigger (frontend-deps.js:2)
```
when input is not empty. According to [this SO thread](https://stackoverflow.com/a/46941547), the fix is to supply such date through the picker initialization method.

Reference: https://github.com/tempusdominus/bootstrap-4/issues/39.


bootstrap4/.js/src/test/scala/io/udash/bootstrap/datepicker/UdashDatePickerTest.scala
```diff
@@ -338,3 +375,9 @@ private trait JQueryDatePickerExt extends JQuery {
   def datetimepicker(function: String): js.Any = js.native
   def datetimepicker(option: String, value: js.Any): JQueryDatePickerExt = js.native
 }
+
+@js.native
+private trait MomentDate extends js.Any {
+  /** @return picker input string (currently set date formatted according to provided `format` string)  */
```
ddworak: Is there a doc link?

bootstrap4/.js/src/main/scala/io/udash/bootstrap/datepicker/UdashDatePicker.scala
```diff
@@ -537,37 +559,53 @@ object UdashDatePicker {
...
+  private val datePickerMutationObserver = {
     new MutationObserver((records: js.Array[MutationRecord], _: MutationObserver) => {
-      val addedNodes = records.flatMap(record => for {i <- 0 until record.addedNodes.length} yield record.addedNodes(i))
-      datePickerSetupCallbacks.foreach { case (pickerId, callback) =>
-        if (addedNodes.exists {
-          case element: Element => element.id == pickerId.id || element.querySelector(s"#$pickerId") != null
-          case _ => false
-        }) callback()
+      def mutationHandler(nodesExtractor: MutationRecord => NodeList, callbacks: MMap[ComponentId, () => Unit]): Unit = {
+        val nodes = records.flatMap(record =>
```
ddworak: Alternative approach:
- register on rendered nodes instead of IDs to avoid query selector
- check if registered nodes are equal or children of removed / added nodes via `Node#contains`

bootstrap4/.js/src/main/scala/io/udash/bootstrap/datepicker/UdashDatePicker.scala
```diff
@@ -154,28 +170,27 @@ final class UdashDatePicker private[datepicker](
...
+        ("date", icons.date),
+        ("up", icons.up),
+        ("down", icons.down),
+        ("previous", icons.previous),
+        ("next", icons.next),
+        ("today", icons.today),
+        ("clear", icons.clear),
+        ("close", icons.close)
+      ).foreach { case (icon, style) =>
+        if (style.nonEmpty) dict.update(icon, style.flatMap(_.classNames).distinct.toJSArray)
```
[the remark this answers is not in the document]
bgrochal: According to the [docs](https://tempusdominus.github.io/bootstrap-4/Options/#icons), not all icons are required to be provided. `DatePickerIcons` splitted into `DefaultDatePickerIcons` enum and `CustomDatePickerIcons` class with neat conversion to a JS dict.

bootstrap4/.js/src/test/scala/io/udash/bootstrap/datepicker/UdashDatePickerTest.scala
bgrochal: Internal [`_i` property](https://momentjs.com/guides/#/lib-concepts/internal-properties/) replaced with a call to [`creationData`](https://momentjs.com/docs/#/parsing/creation-data/).

bootstrap4/.js/src/main/scala/io/udash/bootstrap/datepicker/UdashDatePicker.scala
```diff
@@ -445,17 +446,51 @@ object UdashDatePicker {
...
+      val previous: Option[CssStyle] = Option.empty,
+      val next: Option[CssStyle] = Option.empty,
+      val today: Option[CssStyle] = Option.empty,
+      val clear: Option[CssStyle] = Option.empty,
+      val close: Option[CssStyle] = Option.empty
+  ) extends DatePickerIcons {
+    import scala.scalajs.js.JSConverters._
+
+    override val jsDictionary: js.Dictionary[js.Any] = js.Dictionary(
+      Iterator("time", "date", "up", "down", "previous", "next", "today", "clear", "close")
```
ddworak: `DefaultDatePickerIcon.values.iterator.map(_.name)`

PR #540, opened 2020-02-26
i18n API simplified

ddworak:
`translated(key())` was hard to discover. This proposal sets the "reduction" API as default for keys with arguments and adds extension methods for rendering keys with no more arguments to pass.

I'll update docs once we agree on the final shape of this PR. Please feel free to provide further ideas.


i18n/.js/src/test/scala/io/udash/i18n/LocalTranslationProviderTest.scala
```diff
@@ -9,7 +9,7 @@ class LocalTranslationProviderTest extends UdashFrontendTest {
 
   "LocalTranslationProvider" should {
     "provide translations without argument" in {
-      val translator = new LocalTranslationProvider(
+      implicit val translator = new LocalTranslationProvider(
```
bgrochal: Should we migrate `RemoteTranslationProviderTest` similarly?

i18n/.js/src/main/scala/io/udash/i18n/bindings/TranslationModifier.scala
```diff
@@ -11,15 +11,15 @@ import scalatags.JsDom.Modifier
...
   translation: Future[Translated],
   placeholder: Option[Element],
   rawHtml: Boolean
-) extends Modifier with Bindings with CrossLogging {
+) extends Modifier with CrossLogging {
 
   override def applyTo(t: Element): Unit = {
-    val holder: Seq[Node] = t.appendChild(placeholder.getOrElse(emptyStringNode()))
-    translation onCompleteNow {
+    val holder: Seq[Node] = Seq(t.appendChild(placeholder.getOrElse(Bindings.emptyStringNode())))
```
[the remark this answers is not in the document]
ddworak: `DynamicTranslationBinding` already doesn't extend `Bindings`, this was just a way to avoid larger import for a single symbol here.

i18n/.js/src/test/scala/io/udash/i18n/LocalTranslationProviderTest.scala
ddworak: It's not using `getTranslatedString` helper so we fortunately don't need to.

ddworak:
TODOs before merging:
- docs for extension methods
- attempt to share code between dynamic and static translation modifier
- update i18n extension docs in guide

ddworak:
@bgrochal please take a look at final changes before merge

PR #543, opened 2020-03-03
Fix @transientDefault documentation

OndrejSpanel:
When the value is equal to the default value with `@transientDefault`, the parameter **will** be omitted.


bgrochal:
I'd consider porting the fix to the `0.8.x` branch if it should be available in prod after the next Udash Guide deployment.

PR #551, opened 2020-03-21
Avoid linear list usages in routing

ddworak:
* filtering utils appended to created lists
* `.last` calls etc.
* usage of iterators / indexes where possible

### DIFF
--- a/core/.js/src/main/scala/io/udash/view/ViewRenderer.scala
+++ b/core/.js/src/main/scala/io/udash/view/ViewRenderer.scala
@@ -68,7 +68,7 @@ private[udash] class ViewRenderer(rootElement: => Element) {
     * @param pathToAdd views list, which will be added to hierarchy
     */
   def renderView(subPathToLeave: List[View], pathToAdd: List[View]): Unit = {
-    val currentViewsToLeave = findEqPrefix(subPathToLeave, views.toList)
+    val currentViewsToLeave = findEqPrefix(subPathToLeave.iterator, views.iterator)
 
     if (currentViewsToLeave.isEmpty) {
       require(pathToAdd.nonEmpty, "You cannot remove all views, without adding any new view.")
--- a/core/.js/src/main/scala/io/udash/view/ViewRenderer.scala
+++ b/core/.js/src/main/scala/io/udash/view/ViewRenderer.scala
@@ -6,7 +6,6 @@ import io.udash.utils.FilteringUtils._
 import org.scalajs.dom.Element
 
 import scala.collection.mutable
-import scala.scalajs.js
 
 /**
   * ViewRenderer is used to provide mechanism to render nested [[View]] within provided [[rootElement]].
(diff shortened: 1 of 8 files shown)


core/.js/src/main/scala/io/udash/view/ViewRenderer.scala
```diff
@@ -21,24 +22,23 @@ private[udash] class ViewRenderer(rootElement: => Element) {
...
 
-  private def mergeViews(path: List[View]): Option[View] = {
-    if (path.size == 1) {
-      val singleView: View = path.head
-      views.append(singleView)
-      Some(singleView)
-    } else {
-      val lastElement = path.reduceLeft[View]((parent, child) => {
+  private def mergeViews(path: Iterator[View]): Option[View] = {
+    val result = path.nextOpt.toOption
```
mowczare: NIT: `.setup(_.foreach(...))`

utils/src/main/scala/io/udash/utils/FilteringUtils.scala
```diff
@@ -1,31 +1,15 @@
...
-    def _findEqPrefix(newPath: List[T], previousPath: List[T], subPath: List[T]): List[T] = {
-      (newPath, previousPath) match {
-        case (head1 :: tail1, head2 :: tail2) if head1 == head2 => _findEqPrefix(tail1, tail2, subPath :+ head1)
-        case _ => subPath
-      }
-    }
-
-    _findEqPrefix(newPath, previousPath, Nil)
-  }
+  def findEqPrefix[T](newPath: Iterator[T], previousPath: Iterator[T]): Iterator[T] =
```
mowczare: It's no longer `lists prefix`

utils/src/main/scala/io/udash/utils/FilteringUtils.scala
```diff
@@ -1,31 +1,15 @@
...
-      case (head1 :: tail1, head2 :: tail2) if subPath == Nil && head1 == head2 => _findDiffSuffix(tail1, tail2, subPath)
-      case (head1 :: tail1, _ :: tail2) => _findDiffSuffix(tail1, tail2, subPath :+ head1)
-      case (head1 :: tail1, Nil) => _findDiffSuffix(tail1, Nil, subPath :+ head1)
-      case _ => subPath
-    }
-
-    _findDiffSuffix(newPath, previousPath, Nil)
-  }
+  def findDiffSuffix[T](newPath: Iterator[T], previousPath: Iterator[T]): Iterator[T] =
+    newPath.map(_.opt).zipAll(previousPath.map(_.opt), Opt.Empty, Opt.Empty).dropWhile { case (h1, h2) =>
```
mowczare: possible corner-case (didn't check but worth thinking twice or writing a test):
`newPath` with last element `null` and `previousPath` of size `newPath.size - 1` should give different result than former implementation.

ddworak: Not supposed to be null-safe at all, the `.opt` calls are there just to lift and account for different lengths of these iterators.

PR #552, opened 2020-03-26
scalajs-bundler

ddworak:
This PR is documenting an attempt to use bundler and npm dependencies, might be useful for future migration if it ever happens.

Notes:
- Enforcing `.require` in the right place is sometimes hard, when the usage is not dependent on the module object itself (see bootstrap, where most methods are added to jquery).
- Any module which is not really a module and uses globals makes the dependencies (and their ordering) leak to code even more. In our case the Tempus Dominus Bootstrap 4 Datepicker used global jQuery and moment, which would require us to add custom webpack config.
- JS tests can only be run in node, jsdom is installed automatically though to a specified version. Some things are not implemented (e.g. history), some work differently than in the browser (e.g. range inputs).
- Using npm assets via imports requires e.g. css-loader in webpack or copying similar to the things we did with webjars (done here for comparison).

For modular dependencies (Atmosphere, Moment) the usage is very clean and straightforward. Using bundler would also simplify interaction with third-party wrappers e.g. from Scalably Typed.

Overall, I think it's the way to go in future, but the drawbacks, especially in terms of boilerplate code, are outweighing the advantages for a single page app. I recommend revisiting this in further releases.


sequencer:
any progress of this? I wanna port udash to mill. Dependency to native js is a blocker of this. so I wanna add a `scalajs-bundler` equivalent contrib to mill for this porting.

ddworak:
@sequencer I don't expect much to be honest. You can read the description for the reasons I don't see the bundler as a good fit for udash. The only part of udash which requires any JS dependencies is the bootstrap4 wrapper, which eventually is going to be replaced by our pure Scala.js component framework, thus removing need for any native JS. 

Since there's no equivalent of neither JS dependencies nor bundler dependencies in mill, I'd expect that adding them manually to the `index.html` file (which in case of udash-bootstrap4 means ~1-3 lines depending on what you need) might be the simplest way.

sequencer:
Sorry, I'm not familiar with front end world. If all in Scala would be much better.
So may I ask which file should I edit?
I currently encounter the linking error in ScalaJs compiling, I believe it blames to the native js dependency.

PR #559, opened 2020-04-07
Remove Patch#clearsProperty

### LINKED ISSUE #555: Missing documentation for clearsProperty
ScalaDoc is missing documentation for `clearsProperty` parameter:

https://github.com/UdashFramework/udash-core/blob/4cc35fae02d810af5b81c18f3ca0aed67076987b/core/src/main/scala/io/udash/properties/seq/Patch.scala#L6-L14

My guess is it means the sequence is now empty as a result of the change, but I am not sure.

ddworak:
Closes #555

### DIFF
--- a/core/src/main/scala/io/udash/properties/seq/FilteredSeqProperty.scala
+++ b/core/src/main/scala/io/udash/properties/seq/FilteredSeqProperty.scala
@@ -38,7 +38,7 @@ private[properties] final class FilteredSeqProperty[A, ElemType <: ReadablePrope
       val idx = origin.elemProperties.slice(0, patch.idx).count(p => matcher(p.get))
       CrossCollections.replaceSeq(lastValue, idx, removed.size, added)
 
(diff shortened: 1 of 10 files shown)


core/src/main/scala/io/udash/properties/seq/FilteredSeqProperty.scala
```diff
@@ -38,7 +38,7 @@ private[properties] final class FilteredSeqProperty[A, ElemType <: ReadablePrope
       val idx = origin.elemProperties.slice(0, patch.idx).count(p => matcher(p.get))
       CrossCollections.replaceSeq(lastValue, idx, removed.size, added)
 
-      val filteredPatch = Patch[ElemType](idx, removed, added, lastValue.isEmpty)
+      val filteredPatch = Patch[ElemType](idx, removed, added)
```
mereszeta, line 41: inline?

PR #642, opened 2020-08-31
Turn TranslationKeys into case classes

attomir:
Currently `TranslationKey*` classes do not have the benefit of being case classes, so they only have identity equal and their `toString`s look like `io.udash.i18n.TranslationKey0@garbage`. 

This pull request makes all the implementations of `TranslationKey` into case classes. `TranslationKey0` has private subclasses (which cannot be case classes themselves), so ~~for them the appropriate behavior is coded up manually~~ it is split into `TranslationKey0` trait and `SimpleTranslationKey0` case class.


i18n/src/main/scala/io/udash/i18n/TranslationKey.scala
```diff
@@ -97,12 +100,23 @@ object TranslationKey {
     new Untranslatable(key)
 
   private[i18n] final class ReducedTranslationKey(key: String, val argv: Any*) extends TranslationKey0(key) {
+    override def productPrefix: String = "ReducedTranslationKey"
+    override def productArity: Int = 2
+    override def productElement(n: Int): Any = n match {
```
ddworak, line 105: no real preference yet, but have you considered:
```scala
    override def productArity: Int = argv.size + 1
    override def productElement(n: Int): Any = if(n == 0) key else argv(n - 1)
```

One advantage is you wouldn't have to override `toString` :P

i18n/src/main/scala/io/udash/i18n/TranslationKey.scala
```diff
@@ -9,53 +10,55 @@ sealed trait TranslationKey {
   def key: String
 }
 
-sealed class TranslationKey0(override val key: String) extends TranslationKey {
+sealed case class TranslationKey0(key: String) extends AbstractCase with TranslationKey {
```
ddworak, line 13: I guess you could move `AbstractCase` to `TranslationKey` to avoid repetition

i18n/src/main/scala/io/udash/i18n/TranslationKey.scala
```diff
@@ -97,12 +100,23 @@ object TranslationKey {
...
     override def apply()(implicit provider: TranslationProvider, lang: Lang): Future[Translated] = {
       provider.translate(key, argv: _*)
     }
   }
 
   private[i18n] final class Untranslatable(key: String) extends TranslationKey0(key) {
+    override def productPrefix: String = "Untranslatable"
     override def apply()(implicit provider: TranslationProvider, lang: Lang): Future[Translated] =
       Future.successful(Translated(key))
   }
```
ddworak, line 122: TBH I'm not in love with the design (before your changes as well). Maybe we could break things up a little more, make `TranslationKey0` a trait / abstract class and avoid case class overrides? I don't think the `TranslationKey0` constructor is used very often (there are factories) and we could still provide the `apply` for compatibility. WDYT?

i18n/src/main/scala/io/udash/i18n/TranslationKey.scala
attomir: If it works, sure. I thought it only reduces code size when used directly.

i18n/src/main/scala/io/udash/i18n/TranslationKey.scala
attomir: You mean splitting `TranslationKey0` into a trait (from which `ReducedTranslationKey` and `Untranslatable` would inherit) and a case class? Good idea, but now the hard part is: how to name them?

i18n/src/main/scala/io/udash/i18n/TranslationKey.scala
attomir: I did it this way because:

1. I felt that `Product` implementations have all "hardoded" product arity, so that one class definition can have only one arity.
2. I wanted to mimic adding `case`exactly - but then made exception with `toString` as `"ReducedTranslationKey(key1,WrappedArray(1, 2, 3))"` was just too ugly. :p

Both arguments are rather weak, so no clear preference on my part.

i18n/src/main/scala/io/udash/i18n/TranslationKey.scala
```diff
@@ -1,141 +1,134 @@
...
-        case untranslatable: TranslationKey.Untranslatable =>
-          output.writeSimple().writeString(untranslatable.key)
-        case std: TranslationKey =>
-          val data = output.writeList()
-          data.writeElement().writeSimple().writeString(std.key)
-          data.finish()
-      }
-    }
-  )
+object TranslationKey0 extends HasGenCodec[TranslationKey0] {
```
ddworak, line 132: nit: there should be no reason (historically the way macros worked) for this companion to be at the end of the file

i18n/src/main/scala/io/udash/i18n/TranslationKey.scala
```diff
@@ -1,141 +1,134 @@
...
+  private[i18n] final case class ReducedTranslationKey(key: String, argv: Any*) extends TranslationKey0 {
+    override def apply()(implicit provider: TranslationProvider, lang: Lang): Future[Translated] =
       provider.translate(key, argv: _*)
-    }
+    // default toString puts argv inside "WrappedArray()"
+    override def toString(): String = s"$productPrefix($key,${argv.mkString(",")})"
   }
 
-  private[i18n] final class Untranslatable(key: String) extends TranslationKey0(key) {
+  private [i18n] object ReducedTranslationKey {
```
ddworak, line 108: nit: any formatter will remove the space before square bracket

i18n/src/main/scala/io/udash/i18n/TranslationKey.scala
```diff
@@ -1,141 +1,134 @@
...
+    // note: serialization loses information, as argv is converted to strings
```
ddworak, line 109: Since we control all instantiation of `ReducedTranslationKey`s, maybe we could just change `argv` type to `String*` and render this comment useless? I don't think we're particularly afraid of making `.toString` calls too strict.

ddworak (commented):
I like this draft much more. Let's polish the last couple of details and ship it :)

i18n/src/main/scala/io/udash/i18n/TranslationKey.scala
attomir: I did not do it earlier, as `TranslationProvider.translate` still receives `Any*`, so custom implementations could possibly do some fancy number formatting and the like, which string conversion would foil. Still, I agree that always doing the conversion is preferable to losing info on serialization.

i18n/src/test/scala/io/udash/i18n/TranslationKeyTest.scala
```diff
@@ -115,5 +116,21 @@ class TranslationKeyTest extends UdashSharedTest {
...
+      testKey5.toString should be("TranslationKey5(test5)")
+      testKey6.toString should be("TranslationKey6(test6)")
+      testKey7.toString should be("TranslationKey7(test7)")
+      testKey8.toString should be("TranslationKey8(test8)")
+      testKey9.toString should be("TranslationKey9(test9)")
+      testKeyX.toString should be("TranslationKeyX(testX)")
+      testKeyU.toString should be("Untranslatable(testUntranslatable)")
+      testKeyR.toString should be("ReducedTranslationKey(test3,1,two,5)")
+    }
   }
```
ddworak, line 135: What happened to the equality / comparison tests? It's the initial reason for this changeset after all.

attomir: Thought they are superfluous now, as case classes have a solid implementation of those. I can bring them back if you wish.

ddworak: I'd add them as a contract on what works on what doesn't, there are a couple corner cases with args and types hidden from the user.

ddworak (approved):
LGTM, great work!

PR #643, opened 2020-09-01
2.13 default

ddworak:


core/.js/src/main/scala/io/udash/bindings/Bindings.scala
```diff
@@ -363,35 +363,23 @@ object Bindings extends Bindings {
...
-      * Use this to add more events listeners to an attribute (:= always overrides previous binding).
-      * If callback returns true, other listeners which are queued will not be invoked.
-      * If callback returns false, next callback in the queue will be invoked.
-      */
-    def :+=[T <: Event](callback: T => Boolean): Modifier[Element] = {
-      AttrPair(attr, callback, (el: Element, attr: Attr, callback: T => Boolean) => {
+     * Use this to add more events listeners to an attribute (:= always overrides previous binding).
+     * If callback returns true, other listeners which are queued will not be invoked.
+     * If callback returns false, next callback in the queue will be invoked.
+     * Results other than booleans are treated as false - the don't prevent event propagation.
```
bgrochal, line 369: _the_ -> _they_

core/.js/src/main/scala/io/udash/bindings/Bindings.scala
```diff
@@ -363,35 +363,23 @@ object Bindings extends Bindings {
...
+     */
+    def :+=[T <: Event](callback: T => Any): Modifier[Element] =
```
bgrochal, line 371: Although I like your idea, I'm a little bit afraid of regression. Are you sure no one has called the two-argument variant of `:+=` method with `callback` returning `true` and default `stopPropagation` argument value (`false`)?

As an example, please consider the following case where a compiler won't prevent us from breaking client code:
1. Before:
```scala
def test(x: () => Boolean): Unit = println(s"one-arg with ${x()}")
def test(x: () => Any, retVal: Boolean = false): Unit = println(s"two-arg with ${x(); retVal}")

implicit def stringToBoolean(x: String): Boolean = x.toBoolean
test(() => "true")  // => two-arg with false
```
2. After:
```scala
def test(x: () => Any): Unit = println(s"new version with ${x()}")

implicit def stringToBoolean(x: String): Boolean = x.toBoolean
test(() => "true")  // => new version with true
```

bgrochal (approved):
LGTM

bgrochal:
Also, should the maven central badge in `README.md` point to the 2.13 artifact now?

PR #646, opened 2020-09-04
REST anchors workaround

ddworak:


mereszeta (approved):
I don't have any better idea for it to work, so LGTM i guess

PR #649, opened 2020-09-07
Added support diferent sides at bootstrap's auto margin

catap:
bootstrap supported an auto margin like this `ml-sm-auto` and it is impossibly to constract via type safe methods.

Introduced a way.

Source: https://getbootstrap.com/docs/4.0/utilities/spacing/


mereszeta (approved):
LGTM, thanks for the contribution!

PR #656, opened 2020-09-20
FileService: a way to work with file at frontend

catap:
The usecase of this code is something like this: let image that user is making a protonmail or any another application where content inside frontend (=browser storage) some secret user's data that shouldn't be exposed to backend. Secret key for example or anything like that.

This code allows to developer to convert any `Seq[Array[Byte]]` from frontend to URL as simple call `FileService.createURL`, asynchronously convert any uploaded `File` to `Future[Array[Byte]]` as `FileService.asBytesArray`, or to `Array[Byte]` via `FileService.asSyncBytesArray` inside worker.

Unfortunately scalatags doesn't support `download` attribute and I need to make it by hand. I've opened a PR[^1] to introduce it, but it might be a while until it is included to release.

`FileService.asSyncBytesArray` is using `FileReaderSync` that is also missed inside scala-js-dom. I've opened a PR[^2] but it might be a while. Also, this draft API but it is supported by majority of modern browsers[^3].

[^1]: https://github.com/lihaoyi/scalatags/pull/212

[^2]: https://github.com/scala-js/scala-js-dom/pull/424

[^3]: https://caniuse.com/?search=FileReaderSync

### DIFF
--- a/core/.js/src/main/scala/io/udash/utils/FileService.scala
+++ b/core/.js/src/main/scala/io/udash/utils/FileService.scala
@@ -1,0 +1,133 @@
+package io.udash.utils
+
+import com.avsystem.commons.misc.AbstractCase
+
+import java.io.IOException
+import org.scalajs.dom._
+import org.scalajs.dom.raw.Blob
+
+import scala.scalajs.js
+import scala.concurrent.{Future, Promise}
+import scala.scalajs.js.annotation.JSGlobal
+import scala.scalajs.js.typedarray.ArrayBuffer
+import scala.util.Try
+
+@js.native
+@JSGlobal
+private[utils] final class FileReaderSync() extends js.Object {
+  def readAsArrayBuffer(blob: Blob): ArrayBuffer = js.native
+}
+
+final case class CloseableUrl(value: String) extends AbstractCase with AutoCloseable {
+  override def close(): Unit = {
+    URL.revokeObjectURL(value)
+  }
+}
+
+object FileService {
+
+  final val OctetStreamType = "application/octet-stream"
+
+  /**
+   *  Converts specified bytes arrays to string that contains URL
+   *  that representing the array given in the parameter with specified mime-type.
+   *
+   *  Keep in mind that returned URL should be closed.
+   */
+  def createURL(bytesArrays: Seq[Array[Byte]], mimeType: String): CloseableUrl = {
+    import js.typedarray._
+
+    val jsBytesArrays = js.Array[js.Any](bytesArrays.map(_.toTypedArray) :_ *)
+    val blob = new Blob(jsBytesArrays, BlobPropertyBag(mimeType))
+    CloseableUrl(URL.createObjectURL(blob))
+  }
+
+  /**
+   *  Converts specified bytes arrays to string that contains URL
+   *  that representing the array given in the parameter with `application/octet-stream` mime-type.
+   *
+   *  Keep in mind that returned URL should be closed.
+   */
+  def createURL(bytesArrays: Seq[Array[Byte]]): CloseableUrl =
+    createURL(bytesArrays, OctetStreamType)
+
+  /**
+   *  Converts specified bytes array to string that contains URL
+   *  that representing the array given in the parameter with specified mime-type.
+   *
+   *  Keep in mind that returned URL should be closed.
+   */
+  def createURL(byteArray: Array[Byte], mimeType: String): CloseableUrl =
+    createURL(Seq(byteArray), mimeType)
+
+  /**
+   *  Converts specified bytes array to string that contains URL
+   *  that representing the array given in the parameter with `application/octet-stream` mime-type.
+   *
+   *  Keep in mind that returned URL should be closed.
+   */
+  def createURL(byteArray: Array[Byte]): CloseableUrl =
+    createURL(Seq(byteArray), OctetStreamType)
+
+  /**
+   * Asynchronously convert specified part of file to bytes array.
+   */
+  def asBytesArray(file: File, start: Double, end: Double): Future[Array[Byte]] = {
+    import js.typedarray._
+
+    val fileReader = new FileReader()
+    val promise = Promise[Array[Byte]]()
+
+    fileReader.onerror = (e: Event) =>
+      promise.failure(new IOException(e.toString))
+
+    fileReader.onabort = (e: Event) =>
+      promise.failure(new IOException(e.toString))
+
+    fileReader.onload = (_: UIEvent) =>
+      promise.complete(Try(
+        new Int8Array(fileReader.result.asInstanceOf[ArrayBuffer]).toArray
+      ))
+
+    val slice = file.slice(start, end)
+    fileReader.readAsArrayBuffer(slice)
+
+    promise.future
+  }
+
+  /**
+   * Asynchronously convert specified file to bytes array.
+   */
+  def asBytesArray(file: File): Future[Array[Byte]] =
+    asBytesArray(file, 0, file.size)
+
+  /**
+   * Synchronously convert specified part of file to bytes array.
+   *
+   * Because it is using synchronous I/O this API can be used only inside worker.
+   *
+   * This method is using FileReaderSync that is part of Working Draft File API.
+   * Anyway it is supported for majority of modern browsers
+   */
+  def asBytesArraySync(file: File, start: Double, end: Double): Array[Byte] = {
+    import js.typedarray._
+
+    val fileReaderSync = new FileReaderSync()
+    val slice = file.slice(start, end)
+
+    val int8Array = new Int8Array(fileReaderSync.readAsArrayBuffer(slice))
+
+    int8Array.toArray
+  }
+
+  /**
+   * Synchronously convert file to bytes array.
+   *
+   * Because it is using synchronous I/O this API can be used only inside worker.
+   *
+   * This method is using FileReaderSync that is part of Working Draft File API.
+   * Anyway it is supported for majority of modern browsers
+   */
+  def asBytesArraySync(file: File): Array[Byte] =
+    asBytesArraySync(file, 0, file.size)
+}
--- a/guide/guide/.js/src/main/scala/io/udash/web/guide/views/frontend/demos/FileInputDemo.scala
+++ b/guide/guide/.js/src/main/scala/io/udash/web/guide/views/frontend/demos/FileInputDemo.scala
@@ -1,6 +1,7 @@
 package io.udash.web.guide.views.frontend.demos
 
 import io.udash.css.CssView
+import io.udash.utils.FileService
 import io.udash.web.guide.demos.AutoDemo
 import io.udash.web.guide.styles.partials.GuideStyles
 import scalatags.JsDom.all._
@@ -12,14 +13,35 @@ object FileInputDemo extends AutoDemo with CssView {
     import org.scalajs.dom.File
     import scalatags.JsDom.all._
 
-    val acceptMultipleFiles = Property(true)
+    import scala.concurrent.ExecutionContext.Implicits.global
+
+    val acceptMultipleFiles = true.toProperty
     val selectedFiles = SeqProperty.blank[File]
 
     div(
       FileInput(selectedFiles, acceptMultipleFiles)("files"),
       h4("Selected files"),
       ul(repeat(selectedFiles)(file => {
-        li(file.get.name).render
+        val content = Property(Array.empty[Byte])
+
+        FileService.asBytesArray(file.get) foreach { bytes =>
+          content.set(bytes)
+        }
+
+        val name = file.get.name
+        li(showIfElse(content.transform(_.isEmpty))(
+          span(name).render,
+          {
+            val url = FileService.createURL(content.get)
+            val download = a(href := url.value, attr("download") := name)(name)
+            val revoke = a(href := "#", onclick := { () =>
+              content.set(Array.empty[Byte])
+              url.close()
+            })("revoke")
+
+            Seq(download, span(" or "), revoke).render
+          }
+        )).render
       }))
     )
   }.withSourceCode
--- a/guide/guide/.js/src/main/scala/io/udash/web/guide/views/frontend/FrontendFilesView.scala
+++ b/guide/guide/.js/src/main/scala/io/udash/web/guide/views/frontend/FrontendFilesView.scala
@@ -24,6 +24,8 @@ class FrontendFilesView extends View {
     ),
     p("You can find a working demo application in the ", a(href := References.UdashFilesDemoRepo, target := "_blank")("Udash Demos"), " repositiory."),
     h3("Frontend forms"),
+    p(i("FileService"), " is an object that allows to convert ", i("Array[Byte]")," to URL, save it as file from frontend ",
+      " and asynchronously convert ", i("File"), " to ", i("Future[Array[Byte]]"), "."),
     p(i("FileInput"), " is the file HTML input wrapper providing a property containing selected files. "),
     fileInputSnippet,
     p("Take a look at the following live demo:"),


core/.js/src/main/scala/io/udash/utils/FileService.scala
```diff
@@ -0,0 +1,45 @@
...
+import java.io.IOException
+
+import org.scalajs.dom._
+import scalatags.JsDom.all._
+
+import scala.scalajs.js
+import js.JSConverters._
+import js.typedarray._
+
+import scala.concurrent._
```
ddworak, line 12: Wildcard imports are a tool, not the default - `import scala.concurrent.{Future, Promise}` should be sufficient. I'd probably reduce the scope of some of those as well (e.g. by moving `import scalatags.JsDom.all._` inside the `save` method).

core/.js/src/main/scala/io/udash/utils/FileService.scala
```diff
@@ -0,0 +1,45 @@
...
+import scala.util.Try
+
+object FileService {
+  def save(filename: String, bytes: Array[Byte], mimeType: String = "application/octet-stream"): Unit = {
+    val jsBytes = Array[js.Any](bytes.toTypedArray).toJSArray
```
ddworak, line 17: You can avoid the copy by creating a Javascript array directly: `js.Array[js.Any](bytes.toTypedArray)`

core/.js/src/main/scala/io/udash/utils/FileService.scala
```diff
@@ -0,0 +1,45 @@
...
+    val blob = new Blob(jsBytes, BlobPropertyBag(mimeType))
+    val url = URL.createObjectURL(blob)
+    val download = attr("download")
+
+    val anchor = a(href := url, download := filename, style := "display: none")(filename).render
+    document.body.appendChild(anchor)
+    anchor.click()
```
ddworak, line 24: I understand that this is your use case, but I think even more common one could be to render the anchor in the DOM. We should either provide less functionality or more granular one in library code.

core/.js/src/main/scala/io/udash/utils/FileService.scala
```diff
@@ -0,0 +1,45 @@
...
+    document.body.removeChild(anchor)
+    URL.revokeObjectURL(url)
```
ddworak, line 26: You may revoke here before the file is fully downloaded, which will truncate the download result for larger files / slower downloads.

core/.js/src/main/scala/io/udash/utils/FileService.scala
```diff
@@ -0,0 +1,45 @@
...
+    val promise = Promise[Array[Byte]]()
+
+    fileReader.onerror = (e: Event) =>
+      promise.failure(new IOException(e.toString))
+
+    fileReader.onload = (_: UIEvent) =>
+      promise.complete(Try(
+        new Int8Array(fileReader.result.asInstanceOf[ArrayBuffer]).toArray
+      ))
+
```
ddworak, line 40: `onabort` needs to be handled as well

core/.js/src/main/scala/io/udash/utils/FileService.scala
```diff
@@ -0,0 +1,45 @@
...
+    fileReader.readAsArrayBuffer(file)
+
+    promise.future
+  }
+}
```
ddworak, line 45: This file needs some docs - these are not very straightforward APIs and the use case can be hard to deduce, leading to noone finding / using this.

core/.js/src/main/scala/io/udash/utils/FileService.scala
```diff
@@ -0,0 +1,45 @@
...
+import scalatags.JsDom.all._
+
+import scala.scalajs.js
+import js.JSConverters._
+import js.typedarray._
+
+import scala.concurrent._
+import scala.util.Try
+
+object FileService {
```
ddworak, line 15: Could be tested in `guide-selenium` I guess.

ddworak (changes requested):
Thanks for the contribution, it's surely a tricky API use case we could cover for future users. It needs some more love before we ship it.

core/.js/src/main/scala/io/udash/utils/FileService.scala
catap: It is good question when to revoke it to be honest.

core/.js/src/main/scala/io/udash/utils/FileService.scala
catap: It should be done

core/.js/src/main/scala/io/udash/utils/FileService.scala
catap: Very good point, thanks.

catap:
@ddworak I have no idea when to revoke this URL. If you have some, feel free to share.

core/.js/src/main/scala/io/udash/utils/FileService.scala
catap: I have no idea how to test file uploading via selenium :(

core/.js/src/main/scala/io/udash/utils/FileService.scala
catap: I put an example with revoke to guide and added some comments about it. 

But I have no idea how to do it without leaking, and to reduce "leaking" I made a originally hack with `click` because it allows to obtain URL only when an application needs it.

catap:
@ghik, @ddworak any chance to review? ;)

guide/guide/.js/src/main/scala/io/udash/web/guide/views/frontend/demos/FileInputDemo.scala
```diff
@@ -19,7 +23,27 @@ object FileInputDemo extends AutoDemo with CssView {
       FileInput(selectedFiles, acceptMultipleFiles)("files"),
       h4("Selected files"),
       ul(repeat(selectedFiles)(file => {
-        li(file.get.name).render
+        val content = Property(Array.empty[Byte])
+
+        window.setTimeout(() =>
```
ddworak, line 28: Why is this necessary? It seems to work without it. Let's keep the examples as simple as possible.

guide/guide/.js/src/main/scala/io/udash/web/guide/views/frontend/demos/FileInputDemo.scala
```diff
@@ -19,7 +23,27 @@ object FileInputDemo extends AutoDemo with CssView {
...
+
+        val name = file.get.name
+        li(showIfElse(content.transform(_.isEmpty))(
+          span(name).render,
+          {
+            val anchor = FileService.asAnchor(name, content.get)(name).render
+
+            window.setTimeout(() => {
+              content.set(Array.empty[Byte])
+              URL.revokeObjectURL(anchor.href)
```
ddworak, line 41: TBH I'd rather just show explicitly how revoking works (e.g. add a revoke button).

core/.js/src/main/scala/io/udash/utils/FileService.scala
```diff
@@ -0,0 +1,172 @@
...
+   *  that representing the array given in the parameter with specified mime-type.
+   *
+   *  Keep in mind that returned URL should be revoked via `org.scalajs.dom.revokeObjectURL(url)`.
+   */
+  def asURL(bytesArrays: Seq[Array[Byte]], mimeType: String): String = {
+    import js.typedarray._
+
+    val jsBytesArrays = js.Array[js.Any](bytesArrays.map(_.toTypedArray) :_ *)
+    val blob = new Blob(jsBytesArrays, BlobPropertyBag(mimeType))
+    URL.createObjectURL(blob)
```
ddworak, line 89: Maybe we should work on the return type and pack it in a class, which could have e.g. have `url` and `revoke` methods to notify the user of the workflow with these.

core/.js/src/main/scala/io/udash/utils/FileService.scala
```diff
@@ -0,0 +1,172 @@
...
+
+  /**
+   * Convert specified file to InputStream with blocking I/O
+   *
+   * Because it is using synchronous I/O that could potentially this API can be used only inside worker.
+   *
+   * This method is using FileReaderSync that is part of Working Draft File API.
+   * Anyway it is supported for majority of modern browsers
+   */
+  def asInputStream(file: File, bufferSize: Int = 1024): InputStream =
```
ddworak, line 170: If your intention is just to process a file in chunks and you're already resorting to the working drafts, maybe using `Blob.stream()` would be a better option? The availability is not that poor (https://caniuse.com/?search=Blob.stream) and you could still benefit from async processing. I'm not convinced I want to ship synchronous API with input streams here - we'd have to provide more sane alternatives, which can be handled on the library client side more accurately.

ddworak (changes requested):

core/.js/src/main/scala/io/udash/utils/FileService.scala
```diff
@@ -0,0 +1,172 @@
...
+  def asURL(byteArray: Array[Byte], mimeType: String): String =
+    asURL(Seq(byteArray), mimeType)
+
+  /**
+   *  Converts specified bytes array to string that contains URL
+   *  that representing the array given in the parameter with `application/octet-stream` mime-type.
+   *
+   *  Keep in mind that returned URL should be revoked via `org.scalajs.dom.revokeObjectURL(url)`.
+   */
+  def asURL(byteArray: Array[Byte]): String =
```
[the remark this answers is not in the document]
catap: @ddworak it makes call like `asURL(bytes)()` :(

core/.js/src/main/scala/io/udash/utils/FileService.scala
catap: @ddworak it doesn't work on Safari :(

guide/guide/.js/src/main/scala/io/udash/web/guide/views/frontend/demos/FileInputDemo.scala
catap: Honestly? I haven't remember.

Removed.

core/.js/src/main/scala/io/udash/utils/FileService.scala
```diff
@@ -0,0 +1,172 @@
...
+    asURL(Seq(byteArray), OctetStreamType)
+
+  /**
+   * Create an anchor element that on click downloads specified URL as a file with specified name.
+   */
+  def asAnchor(filename: String, url: String): JsDom.TypedTag[Anchor] = {
```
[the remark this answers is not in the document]
catap: Agreed.

Originally my intended was to prevent duplication of `attr("download")` but it is small price.

guide/guide/.js/src/main/scala/io/udash/web/guide/views/frontend/demos/FileInputDemo.scala
catap: Agreed

core/.js/src/main/scala/io/udash/utils/FileService.scala
catap: I do have two reason to not use `Blob.stream`:
 1. it doesn't work at safari and safari mobile and this two browsers are in the supported scope of project that requires it :)
 2. a scala library that is using this api implements `InputStream` and I need to make a converter between `Blob.stream` and `InputStream` anyway.

=> I decided to move by simpler and most universal way.

And I don't think that forcing someone to use worker to huge job like parsing file is bad idea.

I'd love to say that it is good.

core/.js/src/main/scala/io/udash/utils/FileService.scala
catap: What do you thing about brand new `CloseableUrl`?

guide/guide/.js/src/main/scala/io/udash/web/guide/views/frontend/demos/FileInputDemo.scala
```diff
@@ -19,7 +23,27 @@ object FileInputDemo extends AutoDemo with CssView {
       FileInput(selectedFiles, acceptMultipleFiles)("files"),
       h4("Selected files"),
       ul(repeat(selectedFiles)(file => {
-        li(file.get.name).render
+        val content = Property(Array.empty[Byte])
```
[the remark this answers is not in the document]
catap: `FileService.asBytesArray` is async and I need somehow to replace `span` with file name to my link. I have no idea how to do it.

core/.js/src/main/scala/io/udash/utils/FileService.scala
```diff
@@ -0,0 +1,172 @@
...
+
+import scala.scalajs.js
+import scala.concurrent.{Future, Promise}
+import scala.scalajs.js.annotation.JSGlobal
+import scala.scalajs.js.typedarray.ArrayBuffer
+import scala.util.Try
+
+@js.native
+@JSGlobal
+sealed class FileReaderSync() extends js.Object {
```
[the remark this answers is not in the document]
catap: I can't ride off from `class` here because it requires to [call constructor](https://w3c.github.io/FileAPI/#FileReaderSync). Like [FileReader](https://w3c.github.io/FileAPI/#APIASynch).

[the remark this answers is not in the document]
ddworak: I guess it can still be `private final` so we don't ship it in the API.

core/.js/src/main/scala/io/udash/utils/FileService.scala
ddworak: I think it's the right way to do it. Let's do `extends AbstractCase with AutoCloseable` to save some JS size.

core/.js/src/main/scala/io/udash/utils/FileService.scala
ddworak: I understand your motives, but I'm not convinced we should ship such a complex and use-case specific utility here. Let's strip this method and the whole input stream impl from this PR.

guide/guide/.js/src/main/scala/io/udash/web/guide/views/frontend/demos/FileInputDemo.scala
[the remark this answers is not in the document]
ddworak: Sure, but you can move property creation outside of the lambda (e.g. to where other properties are defined, around line 20), so it doesn't have to be created every time `selectedFiles` change.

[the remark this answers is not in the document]
catap: @ddworak but this `Property` is using for this file :) and when I move property outside of lambda... well, it broke an example when you can upload a few files at one time.

core/.js/src/main/scala/io/udash/utils/FileService.scala
catap: @ddworak do you have anything in mind regarding `asBytesArraySync(...): Array[Byte]`?

core/.js/src/main/scala/io/udash/utils/FileService.scala
catap: agreed

core/.js/src/main/scala/io/udash/utils/FileService.scala
ddworak: No, I think that will be cool!

guide/guide/.js/src/main/scala/io/udash/web/guide/views/frontend/demos/FileInputDemo.scala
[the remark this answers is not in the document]
ddworak: Oh I see. I guess it's good enough for the example, handling this properly would require a bit of complex code.

guide/guide/.js/src/main/scala/io/udash/web/guide/views/frontend/demos/FileInputDemo.scala
```diff
@@ -12,14 +13,35 @@ object FileInputDemo extends AutoDemo with CssView {
     import org.scalajs.dom.File
     import scalatags.JsDom.all._
 
+    import scala.concurrent.ExecutionContext.Implicits.global
+
     val acceptMultipleFiles = Property(true)
```
ddworak, line 18: You can remove this and just use `true.toProperty` for file input arg.

guide/guide/.js/src/main/scala/io/udash/web/guide/views/frontend/demos/FileInputDemo.scala
```diff
@@ -12,14 +13,35 @@ object FileInputDemo extends AutoDemo with CssView {
...
+
+        val name = file.get.name
+        li(showIfElse(content.transform(_.isEmpty))(
+          span(name).render,
+          {
+            val url = FileService.createURL(content.get)
+            val download = a(href := url.value, attr("download") := name)(name)
+            val revoke = a(href := "#", onclick := { () =>
+              content.set(Array.empty[Byte])
+              url.close()
```
ddworak, line 39: You could consider doing `content.clearListeners()` if we don't intend it to change anyway.

core/.js/src/main/scala/io/udash/utils/FileService.scala
catap: @ddworak I just tried to read 100 mb file via this way.

It is a bad idea. Very bad idea :)

So, why you think that `InputStream` is bad way?

guide/guide/.js/src/main/scala/io/udash/web/guide/views/frontend/demos/FileInputDemo.scala
catap: @ddworak I'm using this `content` as `showIfElse(content.transform(_.isEmpty))`. I don't understand how I can move here to `clearListeners()`.

core/.js/src/main/scala/io/udash/utils/FileService.scala
catap: `asSyncBytesArray(file, start, end)` looks quite safe for large file

core/.js/src/main/scala/io/udash/utils/FileService.scala
```diff
@@ -0,0 +1,133 @@
...
+
+  /**
+   * Synchronously convert specified part of file to bytes array.
+   *
+   * Because it is using synchronous I/O this API can be used only inside worker.
+   *
+   * This method is using FileReaderSync that is part of Working Draft File API.
+   * Anyway it is supported for majority of modern browsers
+   */
+  def asSyncBytesArray(file: File, start: Double, end: Double): Array[Byte] = {
```
ddworak, line 112: nit: `asBytesArraySync`

core/.js/src/main/scala/io/udash/utils/FileService.scala
```diff
@@ -0,0 +1,133 @@
...
+
+  /**
+   * Synchronously convert file to bytes array.
+   *
+   * Because it is using synchronous I/O this API can be used only inside worker.
+   *
+   * This method is using FileReaderSync that is part of Working Draft File API.
+   * Anyway it is supported for majority of modern browsers
+   */
+  def asSyncBytesArray(file: File): Array[Byte] =
```
ddworak, line 131: and here

core/.js/src/main/scala/io/udash/utils/FileService.scala
ddworak: Don't get me wrong, buffering is obviously essential for synchronous reading, I'm just not convinced Udash users will benefit from this since:
- 90% cases can be covered directly by `FileReader` and
- the other 10% cases can be covered by adding chunking to it (similar to e.g. https://stackoverflow.com/a/28318964)
- I don't want to encourage blocking the main thread

I hope Java stdlib port evolves enough to support such use cases out of the box (for SJS) https://github.com/scala-js/scala-js/tree/master/javalib/src/main/scala/java/io.

ddworak (approved):
LGTM with minor rename, thanks!

catap:
@ddworak pushed renaming.

PR #660, opened 2020-09-29
Added `udash.local` and `guide.udash.local` to guide's vhost.

catap:
Right now for local development anyone should use `127.0.0.2` as host to access to guide.

Unfortunately not all systems route `127.0.0.0/8` to loopback interface, for example macOS.

That forces a developer to hack this code or put `guide.udash.io` to `/etc/hosts` that might create a error in the future when he forget to remove it.


ddworak (approved):
If it helps then I don't see no reason not to :shipit:

PR #661, opened 2020-09-29
Optional query/header/cookie/body parameter support

ghik:


rest/src/main/scala/io/udash/rest/annotations.scala
```diff
@@ -240,6 +241,38 @@ class Cookie(@defaultsToName override val name: String = RestParamTag.paramName)
 class Body(@defaultsToName override val name: String = RestParamTag.paramName)
   extends rpcName(name) with RestParamTag
 
+/**
+  * Like [[Query]] but indicates that the parameter is optional.
```
ddworak, line 245: Let's add some information about what that indication actually does / changes (to all these), so it's more obvious why someone would want to use this.

rest/src/main/scala/io/udash/rest/openapi/OpenApiMetadata.scala
```diff
@@ -226,14 +227,21 @@ final case class OpenApiBodyOperation[T](
...
   val whenAbsentValue: Opt[JsonValue] = whenAbsentInfo.flatMap(_.fallbackValue)
-  val hasFallbackValue: Boolean = whenAbsentInfo.fold(flags.hasDefaultValue)(_.fallbackValue.isDefined)
+  val isOptional: Boolean = optional || whenAbsentInfo.fold(flags.hasDefaultValue)(_.fallbackValue.isDefined)
+
+  @bincompat private[rest] def hasFallbackValue: Boolean = isOptional
 
   def schema(resolver: SchemaResolver, withDefaultValue: Boolean): RefOr[Schema] =
     resolver.resolve(restSchema) |> (s => if (withDefaultValue) s.withDefaultValue(whenAbsentValue) else s)
+
+  @bincompat private[rest] def this(
```
ddworak, line 242: Good call, are we removing these on merge to `master`?

rest/src/main/scala/io/udash/rest/openapi/RestStructure.scala
```diff
@@ -151,8 +151,10 @@ object RestStructure extends AdtMetadataCompanion[RestStructure] {
   ) extends TypedMetadata[T] {
 
     val fallbackValue: Opt[JsonValue] =
-      (whenAbsentInfo.map(_.fallbackValue) orElse defaultValueInfo.map(_.fallbackValue)).flatten
-    val hasFallbackValue: Boolean = fallbackValue.isDefined
+      if (info.optional) Opt.Empty
+      else (whenAbsentInfo.map(_.fallbackValue) orElse defaultValueInfo.map(_.fallbackValue)).flatten
```
ddworak, line 155: The tests need some love. An obvious (and intentionally not used by you here?) refactoring to `whenAbsentInfo.flatMap(_.fallbackValue) orElse defaultValueInfo.flatMap(_.fallbackValue)` passes.

rest/src/main/scala/io/udash/rest/openapi/OpenApiMetadata.scala
ghik: In principle. yes. In practice, it may not happen immediately at merge but later in bulk.

rest/src/main/scala/io/udash/rest/openapi/RestStructure.scala
ghik: Ok, it seems that other mechanisms make it so that this check is not necessary but I'd leave it anyway to express the intent better and make it safe against possible changes in other parts of the macro system.

ddworak (approved):

PR #662, opened 2020-09-29
Remove `case class Url`

catap:
This case class was since beginng of udash: 7c3176f9b7be419b000546864abf9ca7b42046a4 and had interesting remark: `Url wrapper - just for omitting strings.`

This case class hasn't got anything else except string, and all it forces developer to run a lot of `.value` and `Url(...)`.

Let remove it and simplify api!

### DIFF
--- a/auth/.js/src/test/scala/io/udash/auth/AuthApplicationTest.scala
+++ b/auth/.js/src/test/scala/io/udash/auth/AuthApplicationTest.scala
@@ -1,6 +1,6 @@
 package io.udash.auth
 
-import io.udash.Application
+import io.udash.{Application, Url}
 import io.udash.core._
 import io.udash.routing.RoutingRegistry
 import io.udash.testing.AsyncUdashFrontendTest
@@ -28,8 +28,8 @@ class AuthApplicationTest extends AsyncUdashFrontendTest with AuthTestUtils with
       case "/s3" => ThirdState
     }
 
-    override def matchUrl(url: Url): TestStates = url2State(url.value)
-    override def matchState(state: TestStates): Url = Url(state2Url(state))
+    override def matchUrl(url: Url): TestStates = url2State(url)
+    override def matchState(state: TestStates): Url = state2Url(state)
   }
 
   "AuthApplication" should {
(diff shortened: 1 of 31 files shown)


ddworak:
I like your enthusiasm (and in fact most of the implementation), but I don't think the benefits of this PR outweigh the drawbacks. This is mostly a discussion about value classes (vs type aliases), let me sum up my take on this particular case (class :laughing:) though.

A type alias in Scala is just syntactic convenience - there's no added type safety. The issue with this in Udash is when URL method / constructor arguments are mixed with other string parameters, e.g.
```scala
  class Breadcrumb(val name: String, val link: Url)
```
```scala
    case class Link(title: String, url: Url) extends DefaultDropdownItem
```
The compiler would allow passing these arguments in wrong order. Upgrades /  changes become much more complex when you have to verify these manually. Imagine someone changed one of these classes in Udash, there's no way you'd notice on on upgrade before runtime.

Another thing I like about URL as a value class is that it communicates intent in the API. You get URLs from `io.udash.routing.RoutingRegistry#matchState` or `io.udash.Application#currentUrl` and you can pass them to `io.udash.routing.RoutingRegistry#matchUrl` or `io.udash.routing.RoutingEngine#handleUrl` safely. When both values and arguments are `String`s, the is a lot of places things can go wrong.

Of course the URL class could provide more functionality out of the box (validation, easier modification) or even wrap `java.net.URI` etc., but that's a size / complexity tradeoff in which we've chosen to stay on the more pragmatic side so far. There are other utilities in JVM and JS which can handle those complex cases better.

There are a ton of other arguments for and against value classes and, most prominently, wrapping and unwrapping these can be a burden. The intent of Udash is for you not to ever have to. I don't think we can go forward with this change, but I'd be happy to learn about your use case and how we can avoid forcing you to work with `String`s directly.

catap:
@ddworak yes, very good point about type safe. Honestly it is the major point to close this PR.

I also would like to share there that Java/Scala has `java.net.URL` and `java.net.URI` that has some differences.

For example `{` and `}` should be url-encoded in case of URI by RFC 3986 § 2 when URL says that it shouldn't be url-escaped at https://url.spec.whatwg.org  § 1.3

And this differences can create some pain because Safari uses URI almost everywhere when Chrome uses URL that cause an issue like this: https://bugs.chromium.org/p/chromium/issues/detail?id=660384

So, migrating from `String` to something more complex and smart can introduce browser specified bugs.

Anyway, regarding my case, it is pretty simple: `UrlChangeProvider` is using `url: String` at `changeUrl` but `Url` at `currentFragment`.

So, let me close this PR and fix it on one place :)

PR #663, opened 2020-09-30
Switch to use `Url` at `redirectTo` and `changeUrl`

catap:
It looks confusing now because `Application.currentUrl` returns `Url` when `Application.redirectTo(url)` expects `String`.

As result end user should make something like this
```
val urlBefore = application.currentUrl
//some magic here
application.redirectTo(urlBefore.value)
```

And this use case and this `.value` was a cause for https://github.com/UdashFramework/udash-core/pull/662


ddworak (approved):
LGTM, thanks for catching this!

PR #685, opened 2020-10-23
Test fix

ddworak:
There were 2 types of failures in `RpcSerializationTest`:
- empty text when element was found before RPC call was handled
- `StaleElementReferenceException` when element found by ID was rerendered already

Added top-level `eventually` block and updated `sbt-jsdependencies` to match `scalajs-env-selenium` (which introduced some differences in module handling).


mereszeta (approved):
LGTM

PR #687, opened 2020-10-27
API proposal: ReadableProperty#mirror

ddworak:
This is a pattern we've commonly seen in our codebases whenever a mutable `Property` is an argument to some component while we can only provide a unidirectional stream of updates. This is a draft - feel free to propose alternative approaches or straight out reject this ;)

### DIFF
--- a/core/src/main/scala/io/udash/properties/single/ReadableProperty.scala
+++ b/core/src/main/scala/io/udash/properties/single/ReadableProperty.scala
@@ -74,7 +74,7 @@ trait ReadableProperty[+A] {
     new CombinedProperty[A, B, O](this, property, combiner)
 }
 
-final class MirrorProperty[B](origin: ReadableProperty[B]) {
+final class MirrorProperty[B: PropertyCreator](origin: ReadableProperty[B]) {
   private val castable: CastableProperty[B] = PropertyCreator[B].newProperty(origin.get, null)
   private val registration = origin.streamTo(castable, initUpdate = false)(identity)
   def cancel(): Unit = registration.cancel()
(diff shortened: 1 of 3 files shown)


core/src/main/scala/io/udash/properties/single/ReadableProperty.scala
```diff
@@ -65,6 +74,16 @@ trait ReadableProperty[+A] {
     new CombinedProperty[A, B, O](this, property, combiner)
 }
 
+final class MirrorProperty[B: PropertyCreator](origin: ReadableProperty[B]) {
```
mereszeta, line 77: package private maybe?

mereszeta (approved):
LGTM

core/src/main/scala/io/udash/properties/single/ReadableProperty.scala
ddworak: What would the return type of `def mirror` be then? More importantly, I think it would be hard to make the implicits visible by default then.

bgrochal (approved):
Two minors left, LGTM in general 👍

core/src/main/scala/io/udash/properties/single/ReadableProperty.scala
```diff
@@ -65,6 +74,16 @@ trait ReadableProperty[+A] {
...
+  private val castable: CastableProperty[B] = PropertyCreator[B].newProperty(origin.get, null)
+  private val registration = origin.streamTo(castable, initUpdate = false)(identity)
+  def cancel(): Unit = registration.cancel()
```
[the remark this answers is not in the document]
ddworak: Nice catch!

PR #707, opened 2020-11-27
scala-collection-compat 2.3.1

ddworak:


ddworak:
Waiting for Scala 2.12.13 which should may allow removing silencer altogether.

PR #723, opened 2021-01-19
ConnectionStatus should be final

Starzu:


ddworak (approved):
Good to see you @Starzu! LGTM :shipit:

PR #724, opened 2021-01-19
Scala 2.13.4

ddworak:


project/Dependencies.scala
```diff
@@ -16,7 +16,7 @@ object Dependencies {
   val scalaCssVersion = "0.7.0"
 
   val servletVersion = "4.0.1"
-  val avsCommonsVersion = "2.0.0-M16"
+  val avsCommonsVersion = "2.0.0-M18"
```
sebaciv, line 19: 2.0.0-M19

mereszeta: could even be M21

core/.js/src/main/scala/io/udash/view/ViewRenderer.scala
```diff
@@ -61,7 +62,7 @@ private[udash] class ViewRenderer(rootElement: => Element) {
       case c: ContainerView => c.renderChild(None)
       case _ =>
     }
-    views.trimEnd(views.length - unmodifiedViews)
+    views.trimEnd(views.length - unmodifiedViews): @nowarn
```
mereszeta, line 65: why not `dropRightInPlace` instead?

mereszeta (approved):
LGTM with minor

project/Dependencies.scala
ddworak: The slower the review, the further this gets :smile:

core/.js/src/main/scala/io/udash/view/ViewRenderer.scala
ddworak: If I'm not missing anything there's no easy way to provide it for 2.12 without making things ugly for 2.13.

PR #763, opened 2021-03-24
Fix repeat init with empty SeqProperty

Starzu:
Fixes a bug introduced in 1f752ae22d919b3fc4db85997462144b6be98e73. Since then `repeat` does not append any placeholder on init with empty `SeqProperty`. 

Could you publish an updated version ASAP?


ddworak:
Sure, I'll take a look in the morning and publish if there are no issues.

ddworak (approved):

PR #776, opened 2021-04-14
Fix "Key not found" error in `repeatWithIndex`

Starzu:
Fix "Key not found" error in `repeatWithIndex` caused by updates batched in `CallbackSequencer`.

Reproduction in tests.


ddworak (approved):
Nice catch, thanks!

PR #821, opened 2021-07-07
REST server can be now implemented directly as a class, without API trait

ghik:
This does not exactly fall into original Udash REST use cases (RPC-like with server and client) but I added this because it's easy to implement and there are people in our project interested in this feature.


rest/src/test/scala/io/udash/rest/SomeServerApiImpl.scala
```diff
@@ -0,0 +1,9 @@
+package io.udash.rest
+
+import scala.concurrent.Future
+
+final class SomeServerApiImpl {
+  @GET
+  def thingy(param: Int): Future[String] = Future.successful((param - 1).toString)
+}
```
ddworak, line 8: Could you please test the more esoteric cases here? (abstract method, inherited method, `val`, `lazy val`)

rest/src/test/scala/io/udash/rest/raw/ServerImplApiTest.scala
```diff
@@ -0,0 +1,29 @@
...
+import com.avsystem.commons.Promise
+import io.udash.rest.SomeServerApiImpl
+import org.scalactic.source.Position
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.funsuite.AnyFunSuite
+
+class ServerImplApiTest extends AnyFunSuite with ScalaFutures {
+  private val apiImpl = new SomeServerApiImpl
+  private val serverHandle = SomeServerApiImpl.asHandleRequest(apiImpl)
+
```
ddworak, line 12: Missing some sanity check for OpenAPI generation.

rest/src/test/scala/io/udash/rest/SomeServerApiImpl.scala
ghik: All public methods are scanned. It doesn't matter if they're `val`s, `lazy val`s, abstract or concrete. Methods can be excluded with `@ignore` annotation if desired. I'm already using this technique in Dolang a lot.

ddworak: I know, I just meant a more complex / integration example, not thorough testing of corner cases etc.

ghik: I am very confident that this works, mainly because of Dolang usage. I will add some more tests in `scala-commons` for that, there's nothing Udash specific about this.

ddworak: I'm not going to die on that hill, but I'm a fan of testing the basic use-case (blackbox, integrational) a bit anyway even when using a lib - we have no guarantess (in Udash), that the tests you mentioned will always / still exist on an upgrade, so we could detect the biggest issues then. That might be just a personal paranoia, so feel free to merge.

ddworak (approved):

rest/src/test/scala/io/udash/rest/SomeServerApiImpl.scala
ghik: I added a test involving `@POST`, prefix method and a `val`

PR #854, opened 2021-09-01
Allow callback before application is run

ddworak:
Needed to e.g. initialize pre-app mutation observer correctly without jquery.

### DIFF
--- a/core/.js/src/main/scala/io/udash/Application.scala
+++ b/core/.js/src/main/scala/io/udash/Application.scala
@@ -7,7 +7,9 @@ import io.udash.utils.CallbacksHandler
 import io.udash.view.ViewRenderer
 import org.scalajs.dom
 import org.scalajs.dom.raw.EventListenerOptions
-import org.scalajs.dom.{Element, Event, document}
+import org.scalajs.dom.{Element, Event}
+
+import scala.util.Try
 
 /**
  * Root application which is used to start single instance of app.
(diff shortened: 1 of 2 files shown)


core/.js/src/main/scala/io/udash/Application.scala
```diff
@@ -46,15 +46,18 @@ class Application[HierarchyRoot >: Null <: GState[HierarchyRoot] : PropertyCreat
...
    * Starts the application using selectors to find root element. Handles waiting for document to be ready.
    *
-   * @param selectors A DOMString containing one or more selectors to match.
-   *                  This string must be a valid CSS selector string; if it isn't, a native SyntaxError exception is thrown.
-   *                  See https://developer.mozilla.org/en-US/docs/Web/API/Document_object_model/Locating_DOM_elements_using_selectors.
-   * @param callback  Callback ran after the application is started.
+   * @param selectors            A DOMString containing one or more selectors to match.
+   *                             This string must be a valid CSS selector string; if it isn't, a native SyntaxError exception is thrown.
+   *                             See https://developer.mozilla.org/en-US/docs/Web/API/Document_object_model/Locating_DOM_elements_using_selectors.
+   * @param onContentLoaded      Callback ran on the application root element before the application is started.
```
bgrochal, line 52: In order to precisely indicate when the callback gets executed, I'd name it `onDocumentReady` or modify its description accordingly (e.g. _before the application is started but once the DOM document is ready_).

bgrochal (approved):

core/.js/src/main/scala/io/udash/Application.scala
ddworak: Did both.

PR #863, opened 2021-09-24
Fixed usage of servlet's async context which may be reused between requests

ghik:


rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
```diff
@@ -47,27 +47,32 @@ class RestServlet(
 
   override def service(request: HttpServletRequest, response: HttpServletResponse): Unit = {
     val asyncContext = request.startAsync()
+    val completed = new AtomicBoolean(false)
+
+    // need to protect asyncContext from being completed twice because
+    // servlet may reuse the same context instance between subsequent requests (not cool)
```
ddworak, line 53: nit: `reuse` -> `recycle`? that's the most common terminology regarding this behaviour

rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
```diff
@@ -47,27 +47,32 @@ class RestServlet(
 
   override def service(request: HttpServletRequest, response: HttpServletResponse): Unit = {
     val asyncContext = request.startAsync()
+    val completed = new AtomicBoolean(false)
+
+    // need to protect asyncContext from being completed twice because
```
ddworak, line 52: I'd mention that this happens due to timeout being set (or even add a link if you found sth relevant, e.g. https://stackoverflow.com/a/27744537).

rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
```diff
@@ -47,27 +47,32 @@ class RestServlet(
...
+
     asyncContext.setTimeout(handleTimeout.toMillis)
     asyncContext.addListener(new AsyncListener {
       def onComplete(event: AsyncEvent): Unit = ()
-      def onTimeout(event: AsyncEvent): Unit = {
-        writeFailure(response, Opt("server operation timed out"))
-        asyncContext.complete()
-      }
+      def onTimeout(event: AsyncEvent): Unit =
+        completeWith(writeFailure(response, Opt("server operation timed out")))
```
ddworak, line 64: I think in this case the context is completed by the framework (hence the double completion issue).

rest/.jvm/src/test/scala/io/udash/rest/RestServletTest.scala
```diff
@@ -0,0 +1,60 @@
...
+    server.stop()
+    super.afterAll()
+  }
+
+  test("async context protection") {
+    implicit val backend: SttpBackend[Future, Nothing, WebSocketHandler] = DefaultSttpBackend()
+    val client = SttpRestClient[Waiter](s"http://localhost:$port")
+
+    assertThrows[HttpErrorException](Await.result(client.wait(1100), Duration.Inf))
+    assertThrows[HttpErrorException](Await.result(client.wait(1100), Duration.Inf))
```
ddworak, line 58: Please use a different time here and match the exception message for correct time. It's easier to debug when they're different anyway.

ddworak (commented):
Whoa, there be dragons. I guess you didn't use the alternative (checking if async is started) to avoid race condition? We could also consider handling timeouts on our own but that could get even nastier.

rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
ghik: Double completion was happening because I was completing the request here (due to timeout) and then later when the actual request processing finished on the server.
As far as I understand completing the request here does not hurt (although it may be redundant).

ddworak: My comment comes from the fact that not completing the context here before your changes still triggered the bug. You can even see `onComplete` getting called within this listener.

rest/.jvm/src/test/scala/io/udash/rest/RestServletTest.scala
ghik: The timeout is global so I can't differentiate these requests by timeout exception message.

rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
ghik: That's right. Current implementation with condition variable protects against this too, right?

ddworak: Yeah, it seems it's safe to keep that call.

ghik:
I'm failing to see how checking if async context is started would help with recycled context. Would it not be started by then by another request?
This global timeout should be an absolutely last-resort timeout. The proper timeout should be done on per-request basis, preferably with something like Monix which makes it super easy.

ddworak:
I initially got convinced that somehow only the `AsyncContext` instance would be reused, not the `request` object, but I cannot make it work either. Your solution LGTM.

ddworak (approved):

ghik:
Thx a lot @ddworak

ghik:
Oh great, this is failing non-deterministically.

PR #888, opened 2021-11-02
HttpErrorException accepts arbitrary HTTP body

ghik:


rest/src/main/scala/io/udash/rest/raw/RestRequest.scala
```diff
@@ -50,9 +49,13 @@ object RestParameters {
   final val Empty = RestParameters()
 }
 
-final case class HttpErrorException(code: Int, payload: OptArg[String] = OptArg.Empty, cause: Throwable = null)
-  extends RuntimeException(s"HTTP ERROR $code${payload.fold("")(p => s": $p")}", cause) with NoStackTrace {
-  def toResponse: RestResponse = RestResponse.plain(code, payload)
+case class HttpErrorException(code: Int, payload: HttpBody = HttpBody.Empty, cause: Throwable = null)
```
ddworak, line 52: any reason to open a case class for extension?

ddworak (commented):
Can we have a test for "non-plain" case?

rest/src/main/scala/io/udash/rest/raw/RestRequest.scala
ghik: Actually, yes - that would simplify some code in CDM.

ddworak (approved):

PR #920, opened 2022-01-16
BiDirection RPC

catap:
Here a simple idea to introduce the 3rd type of RPC call: bi-diretion.

This RPC can be used as Client and as Server.


catap:
It was really bad idea

ddworak:
No worries, thanks for trying :)

PR #962, opened 2022-03-23
Get rid of $ from generated ComponentId

Starzu:
jQuery doesn't like `$`s in selectors. Probably jQuery wrapper could also replace `$` in selectors with `\\$`, however, it seems like a good idea to generate ids without `$`s. 

image

image

### DIFF
--- a/macros/src/main/scala/io/udash/macros/ComponentIdMacro.scala
+++ b/macros/src/main/scala/io/udash/macros/ComponentIdMacro.scala
@@ -10,7 +10,7 @@ class ComponentIdMacro(val c: blackbox.Context) {
   final def IdObj: Tree = q"io.udash.component.ComponentId"
 
   def impl(): c.Tree = {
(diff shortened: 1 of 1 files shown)


macros/src/main/scala/io/udash/macros/ComponentIdMacro.scala
```diff
@@ -10,7 +10,7 @@ class ComponentIdMacro(val c: blackbox.Context) {
   final def IdObj: Tree = q"io.udash.component.ComponentId"
 
   def impl(): c.Tree = {
-    val fqn = Iterator.iterate(c.internal.enclosingOwner)(_.owner).find(_.isClass).get.fullName.replace('.', '-')
+    val fqn = Iterator.iterate(c.internal.enclosingOwner)(_.owner).find(_.isClass).get.fullName.replace('.', '-').replace("$", "__")
```
ddworak, line 13: I wouldn't mind replacing with `-` as well or removing altogether.

ddworak (changes requested):
Please add relevant test cases to `io.udash.component.ComponentIdTest`

ddworak:
Closing as stale. Feel free to reopen if needed.

PR #1046, opened 2022-12-19
Added callback to input

mereszeta:
This can allow mitigating some problems with two-way bindings

### DIFF
--- a/core/.js/src/main/scala/io/udash/bindings/inputs/Input.scala
+++ b/core/.js/src/main/scala/io/udash/bindings/inputs/Input.scala
@@ -10,9 +10,10 @@ import scala.concurrent.duration.{Duration, DurationInt}
 /** Abstraction for HTML input tags. */
 private[bindings] abstract class Input(inputType: String) {
   /**
-   * @param value          Property to bind.
-   * @param debounce       Property update timeout after input changes.
-   * @param inputModifiers Additional Modifiers, don't use modifiers on value, onchange and onkeyup attributes.
+   * @param value               Property to bind.
+   * @param debounce            Property update timeout after input changes.
+   * @param onPropertyUpdated   Callback that's being called when property is updated
+   * @param inputModifiers      Additional Modifiers, don't use modifiers on value, onchange and onkeyup attributes.
    * @return HTML input with bound Property, applied modifiers and nested options.
    */
   def apply(value: Property[String], debounce: Duration = 20 millis, onPropertyUpdated: String => Unit = _ => ())(inputModifiers: Modifier*): InputBinding[JSInput] =
--- a/core/.js/src/main/scala/io/udash/bindings/inputs/Input.scala
+++ b/core/.js/src/main/scala/io/udash/bindings/inputs/Input.scala
@@ -7,26 +7,26 @@ import scalatags.JsDom.all._
 
 import scala.concurrent.duration.{Duration, DurationInt}
 
-/** Abstraction for HTML input tags.*/
+/** Abstraction for HTML input tags. */
 private[bindings] abstract class Input(inputType: String) {
   /**
-    * @param value Property to bind.
-    * @param debounce Property update timeout after input changes.
-    * @param inputModifiers Additional Modifiers, don't use modifiers on value, onchange and onkeyup attributes.
-    * @return HTML input with bound Property, applied modifiers and nested options.
-    */
-  def apply(value: Property[String], debounce: Duration = 20 millis)(inputModifiers: Modifier*): InputBinding[JSInput] =
+   * @param value          Property to bind.
+   * @param debounce       Property update timeout after input changes.
+   * @param inputModifiers Additional Modifiers, don't use modifiers on value, onchange and onkeyup attributes.
+   * @return HTML input with bound Property, applied modifiers and nested options.
+   */
+  def apply(value: Property[String], debounce: Duration = 20 millis, onPropertyUpdated: String => Unit = _ => ())(inputModifiers: Modifier*): InputBinding[JSInput] =
     new InputBinding[JSInput] {
       private val element = input(
         inputModifiers, tpe := inputType,
-        nestedInterceptor(new InputModifier(value, Some(debounce)))
+        nestedInterceptor(new InputModifier(value, Some(debounce), onPropertyUpdated))
       ).render
 
       override def render: JSInput = element
     }
 
-  private class InputModifier(property: Property[String], debounce: Option[Duration])
-    extends TextInputsModifier(property, debounce)  {
+  private class InputModifier(property: Property[String], debounce: Option[Duration], onPropertyUpdated: String => Unit = _ => ())
+    extends TextInputsModifier(property, debounce, onPropertyUpdated) {
 
     override def elementValue(t: Element): String =
       t.asInstanceOf[JSInput].value
--- a/core/.js/src/main/scala/io/udash/bindings/inputs/Input.scala
+++ b/core/.js/src/main/scala/io/udash/bindings/inputs/Input.scala
@@ -10,24 +10,35 @@ import scala.concurrent.duration.{Duration, DurationInt}
 /** Abstraction for HTML input tags. */
 private[bindings] abstract class Input(inputType: String) {
   /**
-   * @param value               Property to bind.
-   * @param debounce            Property update timeout after input changes.
-   * @param onPropertyUpdated   Callback that's being called when property is updated
-   * @param inputModifiers      Additional Modifiers, don't use modifiers on value, onchange and onkeyup attributes.
+   * @param value                       Property to bind.
+   * @param debounce                    Property update timeout after input changes.
+   * @param onInputElementEventReceived Callback that's executed when `Input` element receives one of following events:
+   *                                    `Input`, `Change`, `KeyUp`, `Paste` and element value is different than property value.
+   * @param inputModifiers              Additional modifiers. Don't use modifiers on value, onchange and onkeyup attributes
+   *                                    as they are used internally to sync property value with element value.
    * @return HTML input with bound Property, applied modifiers and nested options.
    */
-  def apply(value: Property[String], debounce: Duration = 20 millis, onPropertyUpdated: String => Unit = _ => ())(inputModifiers: Modifier*): InputBinding[JSInput] =
+  def apply(
+    value: Property[String] = Property(""),
+    debounce: Duration = 20 millis,
+    onInputElementEventReceived: String => Unit = _ => ()
+  )(
+    inputModifiers: Modifier*
+  ): InputBinding[JSInput] =
     new InputBinding[JSInput] {
       private val element = input(
         inputModifiers, tpe := inputType,
-        nestedInterceptor(new InputModifier(value, Some(debounce), onPropertyUpdated))
+        nestedInterceptor(new InputModifier(value, debounce, onInputElementEventReceived))
       ).render
 
       override def render: JSInput = element
     }
 
-  private class InputModifier(property: Property[String], debounce: Option[Duration], onPropertyUpdated: String => Unit = _ => ())
-    extends TextInputsModifier(property, debounce, onPropertyUpdated) {
+  private class InputModifier(
+    property: Property[String],
+    debounce: Duration,
+    onInputElementEventReceived: String => Unit = _ => ()
+  ) extends TextInputsModifier(property, debounce, onInputElementEventReceived) {
 
     override def elementValue(t: Element): String =
       t.asInstanceOf[JSInput].value
--- a/core/.js/src/main/scala/io/udash/bindings/inputs/Input.scala
+++ b/core/.js/src/main/scala/io/udash/bindings/inputs/Input.scala
@@ -14,9 +14,9 @@ private[bindings] abstract class Input(inputType: String) {
    * @param debounce            Property update timeout after input changes.
    * @param onInputElementEvent Callback that's executed when `Input` element receives one of following events:
    *                            `Input`, `Change`, `KeyUp`, `Paste` and element value is different than property value.
-   *                            Can be used to mimic unidirectional data flow for components based on this `Input`.
-   * @param inputModifiers      Additional modifiers. Don't use modifiers on value, onchange and onkeyup attributes
-   *                            as they are used internally to sync property value with element value.
+   *                            Can be used to build unidirectional data flow component on top of `Input` component.
+   * @param inputModifiers      Additional modifiers. Attributes: `tpe`, `value`, `onkeyup`, `onchange`, `onpaste`, `oninput`
+   *                            are ignored as they are overwritten internally.
    * @return HTML input with bound Property, applied modifiers and nested options.
    */
   def apply(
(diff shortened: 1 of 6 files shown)


core/.js/src/test/scala/io/udash/bindings/inputs/TextAreaTest.scala
```diff
@@ -134,5 +143,78 @@ class TextAreaTest extends AsyncUdashFrontendTest {
       r.value should be("qaz")
       r2.value should be("asd")
     }
+
+    "run callback on state changes" in {
+      val p = Property[String]("ABC")
+      val result = Property("")
```
ddworak, line 149: using a `Property` here makes the test more vague - a `var` should suffice

core/.js/src/main/scala/io/udash/bindings/inputs/Input.scala
```diff
@@ -7,26 +7,27 @@ import scalatags.JsDom.all._
...
+/** Abstraction for HTML input tags. */
 private[bindings] abstract class Input(inputType: String) {
   /**
-    * @param value Property to bind.
-    * @param debounce Property update timeout after input changes.
-    * @param inputModifiers Additional Modifiers, don't use modifiers on value, onchange and onkeyup attributes.
-    * @return HTML input with bound Property, applied modifiers and nested options.
-    */
-  def apply(value: Property[String], debounce: Duration = 20 millis)(inputModifiers: Modifier*): InputBinding[JSInput] =
+   * @param value               Property to bind.
```
ddworak, line 13: My IDE adds further changes on automatic reformatting

image

core/.js/src/main/scala/io/udash/bindings/inputs/Input.scala
```diff
@@ -7,26 +7,27 @@ import scalatags.JsDom.all._
...
+   * @param debounce            Property update timeout after input changes.
+   * @param onPropertyUpdated   Callback that's being called when property is updated
```
ddworak, line 15: These docs are not conveying the difference between property update and input value update. I'm actually considering if `onPropertyUpdated` is what you want to achieve here long term, or we should use another name. Note that this is an externally passed property, you're not executing the callback when someone changes the value outside of the input.

nit: why is it `being called` and not just `called`?
nit: add missing full stop
nit: you can avoid repetition by changing `called` to `executed`

core/.js/src/main/scala/io/udash/bindings/inputs/Input.scala
```diff
@@ -7,26 +7,27 @@ import scalatags.JsDom.all._
...
+   * @param inputModifiers      Additional Modifiers, don't use modifiers on value, onchange and onkeyup attributes.
```
ddworak, line 16: can we improve this comment? add full stop before `don't` and add some explanation

core/.js/src/main/scala/io/udash/bindings/inputs/Input.scala
```diff
@@ -7,26 +7,27 @@ import scalatags.JsDom.all._
...
+   * @return HTML input with bound Property, applied modifiers and nested options.
+   */
+  def apply(value: Property[String], debounce: Duration = 20 millis, onPropertyUpdated: String => Unit = _ => ())(inputModifiers: Modifier*): InputBinding[JSInput] =
```
ddworak, line 19: nit: line too long

core/.js/src/main/scala/io/udash/bindings/inputs/Input.scala
```diff
@@ -7,26 +7,27 @@ import scalatags.JsDom.all._
...
-        nestedInterceptor(new InputModifier(value, Some(debounce)))
+        nestedInterceptor(new InputModifier(value, Some(debounce), onPropertyUpdated))
       ).render
 
       override def render: JSInput = element
     }
 
-  private class InputModifier(property: Property[String], debounce: Option[Duration])
-    extends TextInputsModifier(property, debounce)  {
+  private class InputModifier(property: Property[String], debounce: Option[Duration], onPropertyUpdated: String => Unit = _ => ())
```
ddworak, line 29: nit: line too long

core/.js/src/main/scala/io/udash/bindings/inputs/TextArea.scala
```diff
@@ -7,25 +7,26 @@ import scalatags.JsDom.all._
...
   /**
-    * @param value Property to bind.
-    * @param debounce Property update timeout after input changes.
-    * @param textareaModifiers Additional Modifiers, don't use modifiers on value, onchange and onkeyup attributes.
-    * @return HTML textarea with bound Property, applied modifiers and nested options.
-    */
-  def apply(value: Property[String], debounce: Duration = 20 millis)(textareaModifiers: Modifier*): InputBinding[TextArea] =
+   * @param value             Property to bind.
+   * @param debounce          Property update timeout after input changes.
+   * @param onPropertyUpdated   Callback that's being called when property is updated.
```
ddworak, line 15: wrong formatting, please also apply comments from `Input`

core/.js/src/main/scala/io/udash/bindings/inputs/TextInputsModifier.scala
```diff
@@ -7,14 +7,22 @@ import org.scalajs.dom._
 import scala.concurrent.duration.Duration
 
 /** Template of binding for text inputs. */
-private[bindings] abstract class TextInputsModifier(property: Property[String], debounce: Option[Duration]) extends Binding {
+private[bindings] abstract class TextInputsModifier(property: Property[String], debounce: Option[Duration], onPropertyUpdated: String => Unit) extends Binding {
```
ddworak, line 10: `debounce` is never `None`

core/.js/src/main/scala/io/udash/bindings/inputs/TextInputsModifier.scala
```diff
@@ -7,14 +7,22 @@ import org.scalajs.dom._
...
   def elementValue(t: Element): String
   def setElementValue(t: Element, v: String): Unit
   def setElementKeyUp(t: Element, callback: KeyboardEvent => Unit): Unit
   def setElementOnChange(t: Element, callback: Event => Unit): Unit
   def setElementOnInput(t: Element, callback: Event => Unit): Unit
   def setElementOnPaste(t: Element, callback: Event => Unit): Unit
 
+  private def updatePropertyValueForElement(element: Element): Unit = {
+    val value: String = elementValue(element)
```
ddworak, line 19: nit: type redundant?

core/.js/src/test/scala/io/udash/bindings/inputs/InputTest.scala
```diff
@@ -162,4 +173,77 @@ class InputTest extends AsyncUdashFrontendTest {
...
   }
+
+  "run callback on state changes" in {
+    val p = Property[String]("ABC")
+    val result = Property("")
+    val input = TextInput(p, 0 millis, result.set(_))()
+    val inputEl = input.render
+
+    inputEl.value = "ABCD"
+    inputEl.onchange(new Event("change"))
```
ddworak, line 184: Previous tests were sending out `null` - is it not enough anymore?

Bonus points for finding some better way of testing input change. If not possible, please extract changing input value to some local extension method.

core/.js/src/main/scala/io/udash/bindings/inputs/TextArea.scala
```diff
@@ -7,25 +7,26 @@ import scalatags.JsDom.all._
 
 import scala.concurrent.duration.{Duration, DurationInt}
 
-/** Simple HTML text area with bound Property.*/
+/** Simple HTML text area with bound Property. */
 object TextArea {
```
ddworak, line 11: What about other `InputBinding`s?

ddworak (changes requested):

core/.js/src/test/scala/io/udash/bindings/inputs/InputTest.scala
mereszeta: It is, but it's clearer to me than sending null

ddworak: Cool, but still please add some local extension method or even just a private method to denoise this code.

core/.js/src/test/scala/io/udash/bindings/inputs/TextAreaTest.scala
```diff
@@ -134,5 +143,76 @@ class TextAreaTest extends AsyncUdashFrontendTest {
...
+      val input = TextArea(p, 0 millis, result = _)()
+      val inputEl = input.render
+
+      inputEl.value = "ABCD"
+      inputEl.onchange(new Event("change"))
+      result should be("ABCD")
+
+      inputEl.value = "ABC"
+      inputEl.onchange(new Event("change"))
+
```
ddworak, line 159: inconsistent line breaks

core/.js/src/main/scala/io/udash/bindings/inputs/Input.scala
```diff
@@ -7,26 +7,38 @@ import scalatags.JsDom.all._
...
-    * @param value Property to bind.
-    * @param debounce Property update timeout after input changes.
-    * @param inputModifiers Additional Modifiers, don't use modifiers on value, onchange and onkeyup attributes.
-    * @return HTML input with bound Property, applied modifiers and nested options.
-    */
-  def apply(value: Property[String], debounce: Duration = 20 millis)(inputModifiers: Modifier*): InputBinding[JSInput] =
+   * @param value                       Property to bind.
+   * @param debounce                    Property update timeout after input changes.
+   * @param onInputElementEventReceived Callback that's executed when `Input` element receives one of following events:
+   *                                    `Input`, `Change`, `KeyUp`, `Paste` and element value is different than property value.
```
ddworak, line 16: Maybe let's skip `Received`.

The doc is not very precise - when the callback is executed, the property value is no longer different.

I still need a sentence about when this would be used.

core/.js/src/main/scala/io/udash/bindings/inputs/Input.scala
```diff
@@ -7,26 +7,38 @@ import scalatags.JsDom.all._
...
+   * @param inputModifiers              Additional modifiers. Don't use modifiers on value, onchange and onkeyup attributes
+   *                                    as they are used internally to sync property value with element value.
+   * @return HTML input with bound Property, applied modifiers and nested options.
+   */
+  def apply(
+    value: Property[String] = Property(""),
+    debounce: Duration = 20 millis,
+    onInputElementEventReceived: String => Unit = _ => ()
```
ddworak, line 24: nit: trailing commas

core/.js/src/test/scala/io/udash/bindings/inputs/InputTest.scala
```diff
@@ -162,4 +173,77 @@ class InputTest extends AsyncUdashFrontendTest {
...
+        result.get should be("A")
+      }
+    } flatMap { _ =>
+      inputEl.value = "123qweasd"
+      inputEl.onchange(new Event("change"))
+      retrying {
+        result.get should be("123qweasd")
+      }
+    }
+  }
```
ddworak, line 248: you're missing some tests for crucial behaviours, e.g.:
- value is set to the same (this can even happen with typing + debounce)
- different event types (supported and unsupported)

core/.js/src/main/scala/io/udash/bindings/inputs/Input.scala
```diff
@@ -7,26 +7,39 @@ import scalatags.JsDom.all._
...
-    * @param debounce Property update timeout after input changes.
-    * @param inputModifiers Additional Modifiers, don't use modifiers on value, onchange and onkeyup attributes.
-    * @return HTML input with bound Property, applied modifiers and nested options.
-    */
-  def apply(value: Property[String], debounce: Duration = 20 millis)(inputModifiers: Modifier*): InputBinding[JSInput] =
+   * @param value               Property to bind.
+   * @param debounce            Property update timeout after input changes.
+   * @param onInputElementEvent Callback that's executed when `Input` element receives one of following events:
+   *                            `Input`, `Change`, `KeyUp`, `Paste` and element value is different than property value.
+   *                            Can be used to mimic unidirectional data flow for components based on this `Input`.
```
ddworak, line 17: nit: does it mimic it or is it actually unidirectional

core/.js/src/main/scala/io/udash/bindings/inputs/Input.scala
```diff
@@ -7,26 +7,39 @@ import scalatags.JsDom.all._
...
+   * @param inputModifiers      Additional modifiers. Don't use modifiers on value, onchange and onkeyup attributes
+   *                            as they are used internally to sync property value with element value.
```
ddworak, line 19: This part of docs is unclear, I'd rather have something like "if you pass modifiers on ..., then ...". They'd probably just get ignored, wouldn't they? (also maybe they don't have to and we can remove the comment then)

Also, the number of modifier types is different than the number of event types above, so you missed sth.

core/.js/src/test/scala/io/udash/bindings/inputs/InputTest.scala
ddworak: where is the test for `value set to the same value due to debouncing`?

core/.js/src/main/scala/io/udash/bindings/inputs/Input.scala
mereszeta: I mean, this component is still bidirectional(it has exposed property), but can be used to create unidirectional layer for higher level components

utils/.js/src/test/scala/io/udash/testing/UdashFrontendTest.scala
```diff
@@ -28,4 +32,19 @@ trait FrontendTestUtils {
 
 trait UdashFrontendTest extends UdashSharedTest with FrontendTestUtils
 
-trait AsyncUdashFrontendTest extends AsyncUdashSharedTest with FrontendTestUtils
\ No newline at end of file
+trait AsyncUdashFrontendTest extends AsyncUdashSharedTest with FrontendTestUtils {
+  def waiting(code: => Any, duration: FiniteDuration)(implicit pos: Position): Future[Assertion] = {
```
ddworak, line 36: nit: the callsites would look better if you separated code from duration

core/.js/src/test/scala/io/udash/bindings/inputs/TextAreaTest.scala
```diff
@@ -222,5 +257,21 @@ class TextAreaTest extends AsyncUdashFrontendTest {
...
+      var called = false
+      val input = TextArea(p, 1 seconds, _ => called = true)()
+      val inputEl = input.render
+
+      called should be(false)
+
+      inputEl.changeValue("CBA")
+      inputEl.changeValue("ABC")
+
+      waiting({
```
ddworak, line 272: I think you can test this more thoroughly without adding a new utility:
- improve the callback to append text to a list
- just use retrying with should not be empty and a time limit - you can map over the result and treat failure as a success

should be a matter of the right `PatienceConfig`

mereszeta: Isn't it the same as current solution but more confusing on the first glance? What would be the gain from that?

ddworak (approved):

PR #1058, opened 2023-01-27
Existing RestSchema for sealed hierarchy improvement

sebaciv:
The goal is to better handle case when existing RestSchema already contains right discriminator field when resolving RestSchema for sealed hierarchy case and allow it to retain custom name (see example in test code).

### DIFF
--- a/rest/src/main/scala/io/udash/rest/openapi/RestSchema.scala
+++ b/rest/src/main/scala/io/udash/rest/openapi/RestSchema.scala
@@ -332,11 +332,6 @@ object InliningResolver {
     new InliningResolver().resolve(schema)
 }
 
-object ShallowInliningResolver extends SchemaResolver {
-  def resolve(schema: RestSchema[_]): RefOr[Schema] =
-    schema.name.fold(schema.createSchema(this))(RefOr.ref)
-}
-
 /**
  * An implementation of [[SchemaResolver]] which registers named [[RestSchema]]s and replaces them with a
  * [[https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#referenceObject Reference Object]].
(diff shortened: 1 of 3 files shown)


rest/src/main/scala/io/udash/rest/openapi/RestSchema.scala
```diff
@@ -332,6 +332,11 @@ object InliningResolver {
     new InliningResolver().resolve(schema)
 }
 
+object ShallowInliningResolver extends SchemaResolver {
```
ghik, line 335: IMO make it private and used only internally

rest/src/main/scala/io/udash/rest/openapi/RestStructure.scala
```diff
@@ -67,23 +67,43 @@ object RestStructure extends AdtMetadataCompanion[RestStructure] {
...
+        restSchema.createSchema(ShallowInliningResolver) match {
+          case RefOr.Value(caseSchema) =>
+            caseSchema.properties.getOpt(cfn) match {
+              case Opt(existingDiscriminator) =>
+                if (existingDiscriminator != caseFieldSchema) {
+                  throw new IllegalArgumentException(
+                    s"Cannot materialize schema for ${info.sourceName}, discriminator field conflict"
+                  )
+                }
+                // When provided `restSchema` already contains right discriminator field just return it unchanged
```
ghik, line 99: "right" => "the expected"

rest/src/main/scala/io/udash/rest/openapi/RestStructure.scala
```diff
@@ -67,23 +67,43 @@ object RestStructure extends AdtMetadataCompanion[RestStructure] {
...
+            )
+            case ref => Schema(allOf = List(RefOr(Schema(
+              `type` = DataType.Object,
+              properties = IListMap(cfn -> caseFieldSchema),
+              required = List(cfn)
+            )), ref))
+          }, taggedName)
+        }
+
+        restSchema.createSchema(ShallowInliningResolver) match {
```
ghik, line 90: `ShallowInliningResolver` could be private somewhere close to this code (e.g. private object in `RestStructure`)

ghik (approved):
All good, pls add some comments as these are trickier parts of the framework.

PR #1124, opened 2023-08-28
Jetty 10 with JSR 356 websocket working

ddworak:


rest/src/test/scala/io/udash/rest/RestApiTest.scala
```diff
@@ -67,7 +67,7 @@ trait RestApiTestScenarios extends RestApiTest {
   }
 
   test("prefixed GET") {
-    testCall(_.prefix("", "h0", "q0").subget(0, 1, 2))
+    testCall(_.prefix("p0", "h0", "q0").subget(0, 1, 2))
```
sebaciv, line 70: Empty prefix no longer works here?

ddworak: Default compliance settings do not allow double slash in URL, as defined in the RFC. They can be overriden individually, but I did not think it important for tests here - do you?

sebaciv: Does `@Prefix("")` usages break in any way? If not - it's OK

ddworak: `io.udash.rest.RestTestApi#transparentPrefix` is not broken. Empty prefix does not mean an empty URL part (double slash), right?

PR #1139, opened 2023-09-25
scalajs-dom 2.7.0

ddworak:


core/.js/src/main/scala/io/udash/utils/FileUploader.scala
```diff
@@ -20,20 +20,26 @@ class FileUploader(url: Url) {
...
   def uploadFile(
-    fieldName: String, file: File, extraData: Map[js.Any, js.Any] = Map.empty, additionalRequestHeaders: Map[RequestName, RequestValue] = Map.empty
+    fieldName: String, file: File, extraData: Map[String, String | Blob] = Map.empty, additionalRequestHeaders: Map[RequestName, RequestValue] = Map.empty
   ): ReadableModelProperty[FileUploadModel] =
     upload(fieldName, Seq(file), extraData = extraData, additionalRequestHeaders = additionalRequestHeaders)
 
   /** Uploads provided `files` in a field named `fieldName` with optional additional request headers. */
   def upload(
-    fieldName: String, files: Seq[File], extraData: Map[js.Any, js.Any] = Map.empty, additionalRequestHeaders: Map[RequestName, RequestValue] = Map.empty
+    fieldName: String, files: Seq[File], extraData: Map[String, String | Blob] = Map.empty, additionalRequestHeaders: Map[RequestName, RequestValue] = Map.empty
```
mereszeta, line 29: Should we add possibility of adding blobName as well? According to provided doc default fileName may not be that helpful:
` The default filename for Blob objects is "blob".`

mereszeta (approved):
Checked release notes, none of the changes besides form data are applicable

core/.js/src/main/scala/io/udash/utils/FileUploader.scala
ddworak: This functionality was not previously exposed by the utility either and there is no matching support on the servlet side either. We can follow up with new feature when needed.

PR #1174, opened 2024-01-16
Possibility to define OpenAPI tags description using annotations

sebaciv:
Add a possibility to define OpenAPI tags with descriptions using annotations in the REST framework

### DIFF
--- a/rest/src/main/scala/io/udash/rest/openapi/OpenApiMetadata.scala
+++ b/rest/src/main/scala/io/udash/rest/openapi/OpenApiMetadata.scala
@@ -54,12 +54,13 @@ final case class OpenApiMetadata[T](
 
   // collect all tags
   private lazy val openApiTags: List[Tag] = {
-    def createTags(method: OpenApiMethod[_]): List[Tag] =
-      method.groupAnnot.fold[List[Tag]](Nil) { group =>
-        List(method.tagAdjusters.foldLeft(Tag(group.groupName))({ case (tag, adjuster) => adjuster.adjustTag(tag) }))
+    def createTag(method: OpenApiMethod[_]): Opt[Tag] =
+      method.groupAnnot.map { group =>
+        method.tagAdjusters.foldLeft(Tag(group.groupName))({ case (tag, adjuster) => adjuster.adjustTag(tag) })
       }
 
-    prefixes.flatMap(prefix => createTags(prefix) ++ prefix.result.value.openApiTags) ++ httpMethods.flatMap(createTags)
+    (prefixes.iterator.flatMap(prefix => createTag(prefix).iterator ++ prefix.result.value.openApiTags.iterator) ++
+      httpMethods.iterator.flatMap(createTag)).toList
   }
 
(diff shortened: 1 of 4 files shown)


rest/src/main/scala/io/udash/rest/openapi/OpenApiMetadata.scala
```diff
@@ -52,6 +52,16 @@ final case class OpenApiMetadata[T](
 ) {
   val httpMethods: List[OpenApiOperation[_]] = (gets: List[OpenApiOperation[_]]) ++ customBodyMethods ++ bodyMethods
 
+  // collect all tags
+  private lazy val openApiTags: List[Tag] = {
```
ddworak, line 56: Will this ever not be evaluated (assuming `OpenApiMetadata` is created) to benefit from laziness? Only when the metadata is not accessed at all? How much computation are we actually saving?

rest/src/main/scala/io/udash/rest/openapi/OpenApiMetadata.scala
```diff
@@ -52,6 +52,16 @@ final case class OpenApiMetadata[T](
...
+    def createTags(method: OpenApiMethod[_]): List[Tag] =
```
ddworak, line 57: This method only creates 0 or 1 "tags". I'd suggest renaming and making it return `Opt[Tag]`, as it simplifies the code without any changes needed to callsites.

rest/src/main/scala/io/udash/rest/openapi/OpenApiMetadata.scala
```diff
@@ -52,6 +52,16 @@ final case class OpenApiMetadata[T](
...
+      method.groupAnnot.fold[List[Tag]](Nil) { group =>
+        List(method.tagAdjusters.foldLeft(Tag(group.groupName))({ case (tag, adjuster) => adjuster.adjustTag(tag) }))
+      }
+
+    prefixes.flatMap(prefix => createTags(prefix) ++ prefix.result.value.openApiTags) ++ httpMethods.flatMap(createTags)
```
ddworak, line 62: You know I'm not a fan of concatenating lists and I'd rather use iterators here to avoid intermediate collections and additional traversal.

rest/src/test/scala/io/udash/rest/RestTestApi.scala
```diff
@@ -124,6 +124,7 @@ trait RestTestApi {
   ): Future[String]
```
ddworak, line 125: missing test case for default group name when not aggregated in `describedGroup`

rest/src/main/scala/io/udash/rest/openapi/OpenApiMetadata.scala
sebaciv: It will be evaluated when `OpenApi` instance is created (io.udash.rest.openapi.OpenApiMetadata#openapi is called). I'm not sure if lazy val is necessary here, but on the other hand other Open API classes like Operations (in io.udash.rest.openapi.OpenApiMetadata#operations) are also created on demand when `openapi` method is called

PR #1238, opened 2024-06-20
RestDataWrapperCompanion refactor

sebaciv:
Extract implicit definitions from `RestDataWrapperCompanion` to allow custom wrapper companions creation.

`BaseRestDataWrapperCompanion` can be reused without copying code, currently `RestDataWrapperCompanion` can't be inherited by another abstract base class because it takes `MacroInstances` as a parameter

### DIFF
--- a/rest/src/main/scala/io/udash/rest/RestDataCompanion.scala
+++ b/rest/src/main/scala/io/udash/rest/RestDataCompanion.scala
@@ -5,7 +5,7 @@ import com.avsystem.commons.meta.MacroInstances
 import com.avsystem.commons.misc.ValueOf
 import com.avsystem.commons.rpc.{AsRaw, AsReal}
 import com.avsystem.commons.serialization.{GenCodec, TransparentWrapperCompanion, TransparentWrapping}
-import io.udash.rest.openapi.*
+import io.udash.rest.openapi._
 import io.udash.rest.openapi.RestStructure.NameAndAdjusters
 import io.udash.rest.raw.{HttpBody, JsonValue, PlainValue, RestResponse}
 
--- a/rest/src/main/scala/io/udash/rest/RestDataCompanion.scala
+++ b/rest/src/main/scala/io/udash/rest/RestDataCompanion.scala
@@ -4,9 +4,9 @@ package rest
 import com.avsystem.commons.meta.MacroInstances
 import com.avsystem.commons.misc.ValueOf
 import com.avsystem.commons.rpc.{AsRaw, AsReal}
-import com.avsystem.commons.serialization.{GenCodec, TransparentWrapperCompanion}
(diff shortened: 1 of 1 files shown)


rest/src/main/scala/io/udash/rest/RestDataCompanion.scala
```diff
@@ -109,3 +100,24 @@ abstract class RestDataWrapperCompanion[Wrapped, T](implicit
...
+  * @example
+  * {{{
+  *   case class UserId(id: String) extends AnyVal
+  *   object UserId extends RestDataWrapperCompanion[String, UserId]
+  * }}}
+  */
+abstract class RestDataWrapperCompanion[Wrapped, T](implicit
+  instances: MacroInstances[DefaultRestImplicits, () => NameAndAdjusters[T]]
+) extends TransparentWrapperCompanion[Wrapped, T] with BaseRestDataWrapperCompanion[Wrapped, T] {
+  final override protected lazy val nameAndAdjusters: NameAndAdjusters[T] = instances(DefaultRestImplicits, this).apply()
```
ddworak, line 121: If we need to cache it (I don't remember the internals of `MacroInstances`), we should add a comment here.

rest/src/main/scala/io/udash/rest/RestDataCompanion.scala
```diff
@@ -43,27 +43,18 @@ abstract class RestDataCompanion[T](implicit
...
-  */
-abstract class RestDataWrapperCompanion[Wrapped, T](implicit
-  instances: MacroInstances[DefaultRestImplicits, () => NameAndAdjusters[T]]
-) extends TransparentWrapperCompanion[Wrapped, T] {
-  private def nameAndAdjusters: NameAndAdjusters[T] = instances(DefaultRestImplicits, this).apply()
+  * Base trait for companion objects of wrappers over other data types (i.e. case classes with single field).
+  * Ensures instances of all the REST typeclasses (serialization, schema, etc.) for wrapping type [[T]]
+  * assuming that these instances are available for the wrapped type [[Wrapped]].
+ */
+trait BaseRestDataWrapperCompanion[Wrapped, T] {
```
ddworak, line 55: Can you add an example (as close to your real use case) in tests? Maybe we could find a better name for what this is then. Also, nobody removes it as "refactoring".

sebaciv: It turns out something like this:
```
abstract class MyCustomRestCompanion[T](
  implicit instances: MacroInstances[DefaultRestImplicits, () => NameAndAdjusters[T]]
) extends RestDataWrapperCompanion[String, T]
```
is actually working, so it turns out I don't need these changes :sweat_smile:

PR #1252, opened 2024-07-26
Add @schemaName annotation to customize RestSchema

sebaciv:
Add @schemaName annotation as a convenient way to customize RestSchema name, unlike com.avsystem.commons.serialization.name does NOT change serialized representation and type discriminator value when the class is a part of a sealed hierarchy.

### DIFF
--- a/rest/src/test/scala/io/udash/rest/openapi/RestSchemaTest.scala
+++ b/rest/src/test/scala/io/udash/rest/openapi/RestSchemaTest.scala
@@ -71,6 +71,22 @@ object FullyQualifiedHierarchy extends RestDataCompanionWithDeps[FullyQualifiedN
   case object Baz extends FullyQualifiedHierarchy
 }
 
+@flatten("case")
+sealed trait CustomSchemaNameHierarchy
+object CustomSchemaNameHierarchy extends RestDataCompanion[CustomSchemaNameHierarchy] {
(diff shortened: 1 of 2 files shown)


rest/src/test/scala/io/udash/rest/openapi/RestSchemaTest.scala
```diff
@@ -71,6 +71,22 @@ object FullyQualifiedHierarchy extends RestDataCompanionWithDeps[FullyQualifiedN
...
+object CustomSchemaNameHierarchy extends RestDataCompanion[CustomSchemaNameHierarchy] {
+  // annotation value should be used as schema name, but NOT as type discriminator value
+  @schemaName("CustomSchemaName123")
+  case class CustomSchemaName(str: String) extends CustomSchemaNameHierarchy
+
+  // annotation value should be used as both schema name and type discriminator value
+  @name("CustomName123")
+  case class CustomName(str: String) extends CustomSchemaNameHierarchy
+
+  // @schemaName should "win" with @name as schema name
```
ddworak, line 85: `@name as schema name` or `@name as type dyscriminator value`? When I look at the declaration I'd expect the latter :D

ddworak (changes requested):

ddworak (approved):

PR #1258, opened 2024-08-05
Replace NativeJsonInput/Output with avs-commons

sebaciv:


benchmarks/.js/src/main/scala/io/udash/benchmarks/serialization/SerializationBenchmarks.scala
sebaciv, line 1: Duplicates serialization benchmarks from avs-commons, so I removed whole file

ddworak (approved):

PR #1303, opened 2024-12-11
close jetty connections on monix timeout (or other cancellation)

halotukozak:
A config for jetty client can have idleTimeout property set to 2h (that means each client connection is closed after 2h if no communication is sent over the connection). This behavior may don't match e.g. 60 seconds monix task timeout setting. Long running http requests inside monix task causes the task to cancel without closing underlying jetty client connection session. Jetty client connection pool is then used up and accepts no more requests effectively blocking the communication.


rest/jetty/src/test/scala/io/udash/rest/jetty/CloseStaleJettyConnectionsOnMonixTimeout.scala
```diff
@@ -0,0 +1,82 @@
...
+    val client = JettyRestClient[RestApiWithNeverCounter](
+      client = httpClient,
+      baseUri = server.getConnectors.head |> { case connector: NetworkConnector => s"http://${connector.getHost}:${connector.getLocalPort}" },
+      maxResponseLength = Int.MaxValue, // to avoid unnecessary logs
+      timeout = RequestTimeout,
+    )
+
+    Task
+      .traverse(List.range(0, Connections))(_ => Task.fromFuture(client.neverGet).timeout(CallTimeout).failed)
+      .timeoutTo(Connections * CallTimeout + 500.millis, Task(fail("All connections should have been closed"))) // + 500 millis just in case
```
ddworak, line 58: Isn't this dead code? It only depends on `Task#timeout` to work and should any of the tasks succeed, we'd get a different exception from `.failed`

rest/jetty/src/test/scala/io/udash/rest/jetty/CloseStaleJettyConnectionsOnMonixTimeout.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  object RestApiWithNeverCounter extends DefaultRestApiCompanion[RestApiWithNeverCounter] {
+    final val Impl: RestApiWithNeverCounter = new RestApiWithNeverCounter {
+      override def neverGet: Future[Unit] = {
+        counter.increment()
+        Future.never
+      }
+    }
+  }
+}
```
ddworak, line 82: Does `SttpRestClient` handle this correctly? If it doesn't we need to solve this within this PR. Also consider making the test generic for all implementations. Adding a testcase to `io.udash.rest.RestApiTestScenarios` would be perfect (if you can avoid so much custom setup, which I think you may be able to do, as there are default timeouts and connection limits to work with).

rest/.jvm/src/test/scala/io/udash/rest/SttpRestCallTest.scala
```diff
@@ -5,7 +5,7 @@ import io.udash.rest.raw.HttpErrorException
 import io.udash.rest.raw.RawRest.HandleRequest
 import sttp.client3.SttpBackend
 
-import scala.concurrent.duration._
+import scala.concurrent.duration.*
```
ddworak, line 8: nit: avoid adding redundant changes to unrelated files

rest/src/test/scala/io/udash/rest/RestTestApi.scala
```diff
@@ -94,6 +95,8 @@ case class ErrorWrapper[T](error: T)
 object ErrorWrapper extends HasPolyGenCodec[ErrorWrapper]
 
 trait RestTestApi {
+  final val neverGetCounter = Atomic(0)
```
ddworak, line 98: That's not an obvious way to do it and in production code would probably make this accessible via REST :D I'd move this to `impl()` and either:

1) Remove the explicit `RestTestApi` type from line 182 and at callsite - Scala will infer a more specific type that allows access to this field.
2) Add methods for getting the counter and resetting it without exposing the implementation.


also: add a comment if you can't think of a better name for this field, otherwise it's hard to understand why it's here without reading all callsites

rest/src/test/scala/io/udash/rest/RestTestApi.scala
```diff
@@ -176,12 +179,15 @@ object RestTestApi extends DefaultRestApiCompanion[RestTestApi] {
...
 
-  val Impl: RestTestApi = new RestTestApi {
+  def impl(): RestTestApi = new RestTestApi {
     def trivialGet: Future[Unit] = Future.unit
     def failingGet: Future[Unit] = Future.failed(HttpErrorException.plain(503, "nie"))
     def jsonFailingGet: Future[Unit] = Future.failed(HttpErrorException(503, HttpBody.json(JsonValue(JsonStringOutput.write(ErrorWrapper("nie"))))))
     def moreFailingGet: Future[Unit] = throw HttpErrorException.plain(503, "nie")
-    def neverGet: Future[Unit] = Future.never
+    def neverGet: Future[Unit] = {
+      neverGetCounter.transform(_ + 1)
```
ddworak, line 188: `neverGetCounter.increment()`

rest/src/test/scala/io/udash/rest/RestApiTest.scala
```diff
@@ -1,19 +1,30 @@
...
 import org.scalatest.concurrent.ScalaFutures
 import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.time.{Millis, Seconds, Span}
+
+import scala.concurrent.duration.FiniteDuration
 
 abstract class RestApiTest extends AnyFunSuite with ScalaFutures {
+  protected final val MaxConnections: Int = 1 // to timeout quickly
+  protected final val Connections: Int = 10 // > MaxConnections
+  protected final val CallTimeout: FiniteDuration = 300.millis // << idle timeout
```
ddworak, line 20: if it's supposed to be much smaller than idle timeout, you should add `assume` clause to the test that requires it

rest/src/test/scala/io/udash/rest/RestApiTest.scala
```diff
@@ -23,16 +34,21 @@ abstract class RestApiTest extends AnyFunSuite with ScalaFutures {
...
-        call(RestTestApi.Impl).catchFailures.wrapToTry.futureValue.map(mkDeep)
+        call(impl).catchFailures.wrapToTry.futureValue.map(mkDeep)
     )
 
   def mkDeep(value: Any): Any = value match {
     case arr: Array[_] => IArraySeq.empty[AnyRef] ++ arr.iterator.map(mkDeep)
     case _ => value
   }
+
+  def getNeverGetCounter(): Int = impl.neverGetCounter.get()
```
ddworak, line 45: nit: this naming is not compliant with scala style

rest/src/test/scala/io/udash/rest/RestApiTest.scala
```diff
@@ -1,19 +1,30 @@
...
+import scala.concurrent.duration.FiniteDuration
 
 abstract class RestApiTest extends AnyFunSuite with ScalaFutures {
+  protected final val MaxConnections: Int = 1 // to timeout quickly
+  protected final val Connections: Int = 10 // > MaxConnections
+  protected final val CallTimeout: FiniteDuration = 300.millis // << idle timeout
+
   implicit def scheduler: Scheduler = Scheduler.global
 
+  private val impl: RestTestApi = RestTestApi.impl()
```
ddworak, line 24: I'd consider switching to `def` here to avoid resetting the counter altogether.

ddworak (changes requested):

rest/src/test/scala/io/udash/rest/RestApiTest.scala
halotukozak: I am not sure if it will work. We have to persist the instance between calls and checks

ddworak: I thought about fixture level persistence

halotukozak: I tried, but macro-related shit beat me, so I put `resetCounter` in `beforeEach`

rest/.jvm/src/test/scala/io/udash/rest/SttpRestCallTest.scala
```diff
@@ -3,13 +3,22 @@ package rest
...
-import scala.concurrent.duration._
+import java.net.http.HttpClient
+import java.time.Duration as JDuration
+import scala.concurrent.duration.*
 import scala.concurrent.{Await, Future}
 
 trait SttpClientRestTest extends ServletBasedRestApiTest {
-  implicit val backend: SttpBackend[Future, Any] = SttpRestClient.defaultBackend()
+  implicit val backend: SttpBackend[Future, Any] = HttpClientFutureBackend.usingClient(
+    //like defaultHttpClient but with connection timeout >> CallTimeout
```
ddworak, line 15: nit: avoid abbreviations in such comments, to some people `>>` means strictly larger, to others it may mean flows, equality, whatever

also, this comment could be a scaladoc for `backend` so it can be accessed from other files

rest/jetty/src/test/scala/io/udash/rest/jetty/JettyRestCallTest.scala
```diff
@@ -6,7 +6,10 @@ import io.udash.rest.{RestApiTestScenarios, ServletBasedRestApiTest}
 import org.eclipse.jetty.client.HttpClient
 
 final class JettyRestCallTest extends ServletBasedRestApiTest with RestApiTestScenarios {
-  val client: HttpClient = new HttpClient
+  val client: HttpClient = new HttpClient() {
```
ddworak, line 9: nit: do we need a comment as in other backends / clients?

ddworak (approved):
LGTM with nits

PR #1447, opened 2026-03-10
Add NestedHierarchy schema and tests

halotukozak:
Non-flat (nested) sealed hierarchy schemas does not have discriminator and should be encoded in open-api spec on its own way.

### DIFF
--- a/rest/.jvm/src/test/resources/RestTestApi.json
+++ b/rest/.jvm/src/test/resources/RestTestApi.json
@@ -805,17 +805,20 @@
         "oneOf": [
           {
             "type": "object",
+            "title": "RestEntity",
             "properties": {
               "RestEntity": {
                 "$ref": "#/components/schemas/RestEntity"
               }
             },
+            "additionalProperties": false,
             "required": [
               "RestEntity"
             ]
           },
           {
             "type": "object",
+            "title": "RestOtherEntity",
             "properties": {
               "RestOtherEntity": {
                 "type": "object",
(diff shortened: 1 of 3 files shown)


rest/src/main/scala/io/udash/rest/openapi/RestStructure.scala
```diff
@@ -40,11 +40,20 @@ object RestStructure extends AdtMetadataCompanion[RestStructure] {
...
+            case RefOr.Value(s) if s.`type`.contains(DataType.Object) && s.additionalProperties == AdditionalProperties.Flag(value = true) =>
+              RefOr(s.copy(additionalProperties = AdditionalProperties.Flag(value = false)))
+            case other => other
+          }
+          RefOr(Schema(
+            `type` = DataType.Object,
+            title = c.info.rawName,
+            properties = IListMap(c.info.rawName -> adjustedBase),
+            additionalProperties = AdditionalProperties.Flag(value = false),
+            required = List(c.info.rawName)
```
sebaciv: Nit: missing ,

rest/.jvm/src/test/resources/RestTestApi.json
```diff
@@ -805,17 +805,20 @@
         "oneOf": [
           {
             "type": "object",
+            "title": "RestEntity",
```
sebaciv, line 808: Hm, are you sure this correct ? I don't know the context of your usecase, but it creates more then one schema with the same title, might break some API generators

sebaciv (changes requested):

rest/.jvm/src/test/resources/RestTestApi.json
halotukozak: the open api generator for sure works wrong for nested hierarchies, because it encodes it the same way like the `flatten` ones (but there is no discriminator). This solution at least works for us (maybe I miss sth else)

sebaciv: I think something might go wrong when you add `@title` for RestEntity class

PR #1545, opened 2026-09-09
Cancel the REST handler task when the client disconnects

pkurcx:
## Summary

`RestServlet` left `AsyncListener.onError` empty, so a client abandoning a request mid-flight was ignored while `onTimeout` already cancelled the handler `Task`. This PR cancels the handler on disconnect too, behind a new `cancelOnDisconnect` flag, and hardens the async completion path so that nothing is written into a request the container has already reclaimed.

> **Release note.** `RestServlet` gains a `cancelOnDisconnect` flag, **off by default**. When enabled, the handler `Task` is cancelled as soon as the container reports the client went away (on Jetty 12 this is delivered for HTTP/2 `RST_STREAM`; an aborted HTTP/1.1 request is not detected while the async cycle is idle and stays bounded by `handleTimeout`). Enable it only once handlers are cancellation-safe: an ordinary client abort (browser navigation, client-side timeout, load balancer reset) would otherwise interrupt a side-effecting handler mid-flight. Independently of the flag, `RestServlet` now stops writing once the request was completed by a timeout or a disconnect, so a late-finishing handler no longer surfaces `IllegalStateException` or bleeds into the next request on the connection.

- **Disconnect detection is transport dependent.** On Jetty 12 an HTTP/2 client abort surfaces as `onError` within milliseconds (`RST_STREAM`). An aborted HTTP/1.1 request is never reported while the async cycle is idle, so it stays bounded only by `handleTimeout`. This is documented on the constructor parameter.
- **`onError` completes the async cycle only for a client disconnect.** Jetty delivers the reset as an `EofException`; for that case the servlet completes the cycle itself, which keeps Jetty from running `sendError(500)` and an `ERROR` dispatch through the application's error handler for every client abort. Any other async failure is logged at warn and left to the container's default error dispatch, so a genuine server-side failure still produces a 500 and stays visible to operators.
- **No writes into a recycled request.** `completed` is also set from `onComplete`, so container-initiated completions are covered. The flag is re-checked before the headers, before the body and before every streamed chunk, and a handler failing after the request was already completed is logged at warn rather than as an application error. Verified against the Jetty 12.1.12 `HttpChannelState` / `ContextHandler` source: the nested `HttpChannel` is cached per connection and recycled between requests.
- **`completeWith` always completes the cycle.** `asyncContext.complete()` runs in a `finally`, so a response write that throws (client gone mid-write) no longer leaves the request pinned until Jetty's own timeout.
- **Listener registered before the handler starts.** `setTimeout` / `addListener` now run before the task is scheduled, with the cancelable held in a `SingleAssignCancelable`; a disconnect arriving before registration is no longer lost, and a fast handler can no longer complete the context before `service()` reaches `addListener`.
- **Binary compatibility.** Shims for the class constructor and `apply` keep the old signatures. The four-argument `apply` shim is `private[rest]` like the existing three-argument one, since Scala prefers an overload that needs no defaults and callers would otherwise trip the bincompat analyzer check.

## Tests

New `RestServletDisconnectTest` runs a Jetty server speaking HTTP/1.1 and cleartext HTTP/2 and covers:

- handler cancellation on HTTP/2 abort with the flag on
- with the flag off, a handler with a non-empty body runs to completion after the abort, is not cancelled, and causes no uncaught errors and no warn/error logs
- no container error dispatch after an abort
- an uncancelable handler with a non-empty body finishing after the abort causes no uncaught errors and no warn/error logs
- a streamed response is stopped after the abort, both via cancellation and, with the flag off, via the per-chunk `completed` check
- an aborted HTTP/1.1 request is still cancelled by `handleTimeout`

Synchronisation uses handler-start latches rather than sleeps, and the HTTP/2 connection is warmed with a ping first, because aborting before the h2c upgrade completes closes the HTTP/1.1 connection instead of resetting a stream.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
```diff
@@ -88,8 +136,14 @@ class RestServlet(
...
-        code
-        asyncContext.complete()
+        try {
+          code
+          asyncContext.complete()
+        } catch {
+          // the container completed the cycle concurrently, i.e. the client went away mid-write
+          case e @ (_: IllegalStateException | _: IOException) =>
+            logger.warn("REST request was completed by the container before its response could be written", e)
+        }
```
sebaciv, line 155: `completed` is flipped *before* `code` runs, so a throwing `code` skips
`asyncContext.complete()` — and every later `completeWith` is then a no-op because the
flag is already `true`.

That's reachable by exactly the scenario this PR targets: the client vanishes mid-write,
`writeResponse`/`writeFailure` throws `EofException`, the catch below logs a WARN, and
the cycle is never completed. `onTimeout`'s `completeWith` can't rescue it either. The
request then sits open, pinning request/connection state, until Jetty force-completes it
at `handleTimeout` (30s by default).

The old code had the same hole, but it at least surfaced the exception instead of
swallowing it at WARN.

rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
```diff
@@ -98,13 +152,18 @@ class RestServlet(
       (for {
         restRequest <- Task.fromTry(udashRequest)
         restResponse <- handleRequest(restRequest)
-        _ <- Task(setResponseHeaders(response, restResponse.code, restResponse.headers))
-        _ <- writeResponseBody(response, restResponse)
+        // an uncancelable handler may outlive the request; the response object then already belongs to the container
+        _ <- if (completed.get) Task.unit else
```
sebaciv: This guard samples the flag once, right after `handleRequest` finishes, so it's a TOCTOU
check rather than a guard.

If `onError`/`onTimeout` fires just after the read, it sets `completed` and calls
`asyncContext.complete()` — at which point Jetty recycles the `Request`/`Response` for
the next request on that connection, while `setResponseHeaders` / `writeResponseBody`
are still running against that same `HttpServletResponse`. For a `StreamedRestResponse`
the window isn't an instant, it's the entire duration of the stream, and
`cancelable.cancel()` is asynchronous so it doesn't close the write either.

Two outcomes: status/headers/bytes of the abandoned response bleed into an unrelated
request's response, or an `IllegalStateException` surfaces as
`logger.error("Failed to handle REST request")` for what is a benign disconnect.

The new test doesn't cover this — the `uncancelable` handler sleeps a full second *after*
the abort, so the write always happens well clear of the completion.

Worth hardening: also set `completed` from `onComplete` (so container-initiated
completions are covered, not just the two listener paths), and re-check the flag inside
the write path rather than only once ahead of it.

rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
```diff
@@ -117,7 +176,14 @@ class RestServlet(
...
       }
-      def onError(event: AsyncEvent): Unit = ()
+      // The container reports a failed async cycle here, most importantly the client going away. No response can
+      // reach the client anymore; completing the cycle here keeps the container from running its default error
+      // dispatch (a 500 through the error handler) for what is not an application failure.
+      def onError(event: AsyncEvent): Unit = {
+        logger.debug("REST request failed asynchronously, most likely the client disconnected", event.getThrowable)
+        if (cancelOnDisconnect) cancelable.cancel()
+        completeWith(())
+      }
```
sebaciv, line 180: `onError` isn't raised only for client disconnects — Jetty raises it for any failure of
the async cycle: response write failures, connection idle timeout during async, errors
from a nested dispatch. Treating all of them as "the client is gone" changes two things:

- Previously such a failure fell through to the container's error dispatch and produced
  a 500. Now the servlet unconditionally calls `completeWith(())`, so where the
  connection is still usable the peer gets the default `200` with an empty or truncated
  body instead.
- The throwable is only visible at DEBUG, so operators lose the signal entirely for
  genuine server-side async failures.

Suggestion: log non-`EOFException` throwables at WARN, and/or only self-completing when
the throwable actually indicates the client is gone — falling through to the error
dispatch otherwise.

rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
```diff
@@ -24,6 +24,7 @@ object RestServlet {
   final val DefaultMaxPayloadSize = 16 * 1024 * 1024L // 16MB
   final val CookieHeader = "Cookie"
   final val DefaultStreamingBatchSize = 100
+  final val DefaultCancelOnDisconnect = true
```
sebaciv: This PR is careful about *binary* compatibility (the `@bincompat` overloads), but
defaulting to `true` flips runtime semantics for every existing user.

Any handler that performs side effects — a `POST` that writes to a DB, publishes a
message, charges a card — will now be cancelled mid-flight whenever a client aborts the
stream: browser navigation, a client-side timeout, a load balancer sending
HTTP/2 `RST_STREAM`. Previously the task ran to completion; now it can leave partially
applied effects.

Given the library's compatibility posture, `false` (opt-in) seems the safer default —
cancellation is a real improvement, but it's the kind that users should be able to reach
for deliberately. At minimum this needs a release-note callout.

rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
sebaciv, line 161: `setTimeout` and `addListener` run *after* the task has already been handed to the
scheduler by `.executeAsync.runAsync`, which leaves a startup blind spot now that
cancellation depends on the listener:

- A disconnect reaching Jetty between `request.startAsync()` and `addListener(...)` is
  delivered to no listener at all, so `cancelOnDisconnect` silently does nothing for it.
- Because `.executeAsync` hands the task to a scheduler thread, a fast handler can
  complete the context before the Jetty thread gets here, and then `setTimeout` /
  `addListener` throw `IllegalStateException` out of `service()`. That one is
  pre-existing, but this PR makes correct cancellation depend on the listener being
  registered, which raises the stakes.

Both go away if `setTimeout` + `addListener` move above `runAsync`, with the cancelable
held in a `SingleAssignCancelable` that the listener closes over.

rest/.jvm/src/test/scala/io/udash/rest/RestServletDisconnectTest.scala
```diff
@@ -0,0 +1,178 @@
...
+  "RestServlet" should {
+    "cancel the handler task when an HTTP/2 client aborts the request" in {
+      abortHttp2Request("api/hanging/h2-abort")
+      assert(latch(cancelled, "h2-abort").await(4, TimeUnit.SECONDS))
+    }
+
+    "not cancel the handler task when cancelOnDisconnect is disabled" in {
+      abortHttp2Request("no-cancel-api/hanging/h2-abort-disabled")
+      assert(!latch(cancelled, "h2-abort-disabled").await(2, TimeUnit.SECONDS))
+    }
```
sebaciv: Test leaks a `Task.never` for the rest of the suite: `onError` completes the async cycle, so `onTimeout` never fires to cancel it.

sebaciv (changes requested):

rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
pkurcx: Fixed in a9b390d8: `asyncContext.complete()` now runs in a `finally`, so a throwing write still completes the cycle. The write failure itself is logged at warn; an `IllegalStateException` from `complete()` (the container completed concurrently) at debug.

rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
pkurcx: Fixed in a9b390d8. `onComplete` now sets `completed`, and the flag is re-checked before the headers, before the body and before every streamed chunk (`takeWhile` on the chunk/batch stream, plus the closing `]`). A handler failing once `completed` is set is logged at warn instead of error. The residual window is a single write call racing the container's recycle, which the servlet API gives no way to close. Added streaming scenarios in both modes; the disabled-mode one exercises the per-chunk check rather than cancellation.

rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
pkurcx: Fixed in a9b390d8. `onError` self-completes only for an `EOFException` (Jetty's `EofException` extends it, and `HttpStreamOverHTTP2` wraps a remote reset in one). Anything else is logged at warn with the throwable and only marks the response as no longer ours, so the container's default error dispatch still answers with a 500.

rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
pkurcx: Flipped to `false` in a9b390d8 and documented the reasoning on the constructor parameter. The PR description now carries a release-note callout.

rest/.jvm/src/main/scala/io/udash/rest/RestServlet.scala
pkurcx: Fixed in a9b390d8: `setTimeout` and `addListener` run before the task is scheduled, and the listener closes over a `SingleAssignCancelable` that is assigned afterwards (a cancel before assignment cancels on assignment).

rest/.jvm/src/test/scala/io/udash/rest/RestServletDisconnectTest.scala
pkurcx: Fixed in a9b390d8: the disabled-mode test now uses a handler that answers with a non-empty body a second after the abort and asserts it finished, was not cancelled, and produced no warn/error logs. That also covers the late-write path in that mode.

